PR #35, opened 2017-01-04
Add makemkv profile for controlling makemkv audio/subtitle selection

RandomNinjaAtk:
I am currently using parts of this in my own setup. I've implemented this in my own setup so I could customize the profile used by Makemkv to ensure HD audio is selected when ripping to non-compressed MKV rips. This needs to be thoroughly tested... The part that is untested is using variables to pass the config file path and the augmentation of the makemkvcon command using variables. I currently have this working in my setup strictly using the intended output without any configuration via variables.

### DIFF
--- a/config.sample
+++ b/config.sample
@@ -54,6 +54,12 @@ MINLENGTH="600"
 # DVD's will always default back to the "mkv" mode. If this is set to "backup" then you must also set HandBrake's MAINFEATURE to true. 
 RIPMETHOD="mkv" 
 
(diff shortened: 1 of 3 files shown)


ahnooie:
Thanks for the PR @RandomNinjaAtk  I'll try to test it out later this week.

RandomNinjaAtk:
I've updated my own branch and currently testing the combination of both pull requests with a couple other minor changes to file moves to automate it further. I will submit a new pull request from my branch when its complete! I'm doing my final testing now :D

PR #37, opened 2017-01-06
Add Plex Support

RandomNinjaAtk:
This adds additional configuration to enable usage with Plex, by default this script is built to work with Emby after it's most recent updates. This update enables proper use with Plex, by creating the directories necessary for Extras. This will also enable dual use, by creating the proper files for both Plex and Emby to pull from the same directories, encase one is running both applications at the same time...

Please review my code thoroughly, as I am a bit of a novice when it comes to these things.... I am primarily a Plex user, that is why I made this patch and it would be beneficial to others...

### DIFF
--- a/config.sample
+++ b/config.sample
@@ -30,7 +30,7 @@ RAWPATH="/mnt/media/ARM/raw/"
 
 # Path to final media directory
 # Destination for final file.  Only used for movies that are positively identified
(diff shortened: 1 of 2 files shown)


RandomNinjaAtk:
I've updated my own branch and currently testing the combination of both pull requests with a couple other minor changes to file moves to automate it further. I will submit a new pull request from my branch when its complete! I'm doing my final testing now :D

PR #38, opened 2017-01-07
Added MakeMKV Profile, Plex and Main movie Support

RandomNinjaAtk:
So this is my combined merge, I have tested this myself. It adds support for Plex, MakeMKV profiles and Main Movie. The beauty of this is that it also enhances Emby, to allow users to run both Plex and Emby using the same set of files.

This change will also enable the ARM to guess at which file is the main movie. When moving the files to the Extras/Featurettes sub-folders, the script will now access those files, select the largest file in terms of file size and move it the root of the movie folder. That way when Plex or Emby scans the directories, the movie can be automatically added as it is supposed too!

If you have any questions, let me know. I'm ripping straight to MKV with no transcode, these changes only mainly effect if your not transcoding. If you are transcoding, the MakeMKV Profile will still work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,10 +48,10 @@ See: https://b3n.org/automatic-ripping-machine
     sudo apt install python3 python3-pip
     sudo apt-get install libdvd-pkg
     sudo dpkg-reconfigure libdvd-pkg
-    mkdir -p ~/ARM/raw
-    mkdir -p ~/ARM/Media/Movies
-    mkdir ~/ARM/Media/Music
-    mkdir ~/ARM/Media/Unidentified
+    sudo mkdir -p /mnt/media/ARM/raw
+    sudo mkdir -p /mnt/media/ARM/Media/Movies
+    sudo mkdir /mnt/media/ARM/Media/Music
+    sudo mkdir /mnt/media/ARM/Media/Unidentified
     sudo su
     cd /opt
     git clone https://github.com/RandomNinjaAtk/automatic-ripping-machine.git arm
--- a/README.md
+++ b/README.md
@@ -48,10 +48,7 @@ See: https://b3n.org/automatic-ripping-machine
     sudo apt install python3 python3-pip
     sudo apt-get install libdvd-pkg
     sudo dpkg-reconfigure libdvd-pkg
-    sudo mkdir -p /mnt/media/ARM/raw
-    sudo mkdir -p /mnt/media/ARM/Media/Movies
     sudo mkdir /mnt/media/ARM/Media/Music
-    sudo mkdir /mnt/media/ARM/Media/Unidentified
     sudo su
     cd /opt
     git clone https://github.com/RandomNinjaAtk/automatic-ripping-machine.git arm
--- a/README.md
+++ b/README.md
@@ -34,24 +34,24 @@ See: https://b3n.org/automatic-ripping-machine
 
 ## Install
 
-    sudo su
-    add-apt-repository ppa:heyarje/makemkv-beta
-    add-apt-repository ppa:stebbins/handbrake-releases
-    add-apt-repository ppa:mc3man/xerus-media
-    apt update
-    apt install makemkv-bin makemkv-oss
-    apt install handbrake-cli libavcodec-extra
-    apt install abcde flac imagemagick glyrc cdparanoia
-    apt install at
-    apt install python3 python3-pip
-    apt-get install libdvd-pkg
-    dpkg-reconfigure libdvd-pkg
+    sudo add-apt-repository ppa:heyarje/makemkv-beta
+    sudo add-apt-repository ppa:stebbins/handbrake-releases
+    sudo add-apt-repository ppa:mc3man/xerus-media
+    sudo apt update
+    sudo apt install makemkv-bin makemkv-oss
+    sudo apt install handbrake-cli libavcodec-extra
+    sudo apt install abcde flac imagemagick glyrc cdparanoia
+    sudo apt install at
+    sudo apt install python3 python3-pip
+    sudo apt-get install libdvd-pkg
+    sudo dpkg-reconfigure libdvd-pkg
     mkdir -p ~/ARM/raw
     mkdir -p ~/ARM/Media/Movies
     mkdir ~/ARM/Media/Music
     mkdir ~/ARM/Media/Unidentified
+    sudo su
     cd /opt
-    git clone https://github.com/ahnooie/automatic-ripping-machine.git arm
+    git clone https://github.com/RandomNinjaAtk/automatic-ripping-machine.git arm
     cd arm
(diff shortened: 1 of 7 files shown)


RandomNinjaAtk:
I see it says it's failing on one check, I'm not sure how to fix that using the find command. Maybe if you have a better understanding of how too, feel free to change it. I'm using the code as is and it's working great... That function looks at the size of the files in the Extra/Featurettes directory, grabs the largest file in terms of size, and then moves it to the root of the movie folder....

ahnooie:
Thanks, @RandomNinjaAtk, we can probably just add a shellcheck disable for that line that's causing the build to fail, shellcheck doesn't like ls.   I'll test this out hopefully later this week.

RandomNinjaAtk:
I think my changes to the .abcde.conf file make it more plex/kodi/Emmy friendly. Why the revert?

ahnooie:
Ah, I thought changes to that file were incidental, now that I see what your changes did to various aryistd I think I like your version of .abcde.conf  better.

RandomNinjaAtk:
ahnooie, do you use Skype ? Want to talk in person to better collaborate on what I'm trying to do with the modifications.

README.md
```diff
@@ -34,20 +34,23 @@ See: https://b3n.org/automatic-ripping-machine
 
 ## Install
 
+    sudo apt-get install git
+    sudo apt-get install regionset
+    sudo regionset /dev/sr0
```
ahnooie: This is good, I think there should be a note warning that the region can only be changed 5? times on most DVD players then it gets stuck.  Also might mention if the region is already set there's no need to run it.

data_rip.sh
```diff
@@ -10,15 +10,16 @@ source "$ARM_CONFIG"
...
 	FILENAME=${ID_FS_LABEL}_disc.iso
 
 
 	#dd if=/dev/sr0 of=$DEST/$FILENAME 
 	cat "$DEVNAME" > "$DEST/$FILENAME"
 
 	eject
-
+	
+	chmod 777 -R "$DEST"
```
ahnooie: Some people with a NAS may not want to set it to 777.  Perhaps add an option in the config file, maybe "SET_MEDIA_PERMISSIONS="true" / "false" ?

video_transcode.sh
```diff
@@ -151,6 +155,9 @@ TIMESTAMP=$5
 		echo "***WARNING!*** This will likely leave files in the transcoding directory as there is very likely existing files in the media directory"
         echo "Moving multiple files to emby movie folder" >> "$LOG"
 		mv -n "$DEST/$LABEL.$DEST_EXT" "$MEDIA_DIR/$LABEL.$DEST_EXT"
+		
+		chmod 777 "$MEDIA_DIR" -R
```
ahnooie: See comment on data_rip.sh

video_transcode.sh
```diff
@@ -162,23 +169,56 @@ TIMESTAMP=$5
 		echo '$VIDEO_TYPE is movie, $MAINFEATURE is false, $HAS_NICE_TITLE is true, $EMBY_SUBFOLDERS is true' >> "$LOG"
         echo "Moving multiple files to emby movie subfolders" >> "$LOG"
 		echo "First move main title" >> "$LOG"
-        mkdir -v "$MEDIA_DIR/$LABEL" >> "$LOG"
+        mkdir -p -v "$MEDIA_DIR/$LABEL" >> "$LOG"
 		if [ ! -f "$MEDIA_DIR/$LABEL/$LABEL.$DEST_EXT" ]; then
 			echo "No file found.  Moving \"$DEST/$LABEL.$DEST_EXT to $MEDIA_DIR/$LABEL/$LABEL.$DEST_EXT\"" >> "$LOG"
 			mv -n "$DEST/$LABEL.$DEST_EXT" "$MEDIA_DIR/$LABEL/$LABEL.$DEST_EXT" >> "$LOG"
+			chmod 777 "$MEDIA_DIR/$LABEL" -R
```
ahnooie: See comment on data_rip.sh

video_transcode.sh
```diff
@@ -162,23 +169,56 @@ TIMESTAMP=$5
...
+			
+			# shellcheck disable=SC2086
+       			echo "Sending command: mv -n "\"$DEST/$LABEL/*\"" "\"$MEDIA_DIR/$LABEL/Featurettes/\""" >> "$LOG"
+       			mv -n "${DEST}"/* "$MEDIA_DIR/$LABEL/Featurettes/" >> "$LOG"
+			
+			# Move Largest file to main folder for Plex/Emby/Kodi to detect main movie
+			# shellcheck disable=SC2012
+			ls -S "$MEDIA_DIR/$LABEL/Featurettes/" | head -1 | xargs -I '{}' mv "$MEDIA_DIR/$LABEL/Featurettes/"{} "$MEDIA_DIR/$LABEL/$LABEL.mkv" >> "$LOG"
+			
+			chmod 777 "$MEDIA_DIR/$LABEL" -R
```
ahnooie: See comment on data_rip.sh

video_transcode.sh
```diff
@@ -162,23 +169,56 @@ TIMESTAMP=$5
...
+			
+			# shellcheck disable=SC2086
+      			echo "Sending command: mv -n "\"$DEST/$LABEL/*\"" "\"$MEDIA_DIR/$LABEL/extras/\""" >> "$LOG"
+       			mv -n "${DEST}"/* "$MEDIA_DIR/$LABEL/extras/" >> "$LOG"
+			
+			# Move Largest file to main folder for Plex/Emby/Kodi to detect main movie
+			# shellcheck disable=SC2012
+			ls -S "$MEDIA_DIR/$LABEL/extras/" | head -1 | xargs -I '{}' mv "$MEDIA_DIR/$LABEL/extras/"{} "$MEDIA_DIR/$LABEL/$LABEL.mkv" >> "$LOG"
+			
+			chmod 777 "$MEDIA_DIR/$LABEL" -R
```
ahnooie: See comment on data_rip.sh

data_rip.sh
ahnooie: Also, I might be a bit picky, but per the man page the options (-R) should come before the numeric mode.  E.g.  "chmod -R 777 "$DEST"

README.md
RandomNinjaAtk: So I have fixed everything and made it even better for customizing the file permissions based on your suggestion. I am testing it now but very confident in it. For the README.md, I'll let you handle updating that, you can just revert mine and use my suggestions to help update the original.

RandomNinjaAtk:
This is ready to be merged, I'll let you handle updating the readme, or after you've merged my code, will make updates to it because I will be using the master, and not my fork as the base for my future changes, if I happen to make any more...

PR #58, opened 2017-04-15
Adding chown flag option to change user:group of files

timshelley:
Adding the option to set the user/group of generated files


ahnooie:
This makes sense, thanks for the PR @timshelley

Issue #59, opened 2017-05-18
HDDVDs are being identified as UDF DATA instead of Video


TheFiZi:
Tested with Anchorman and Beowulf HDDVD.

Here is the log for Beowulf:

```
*** Start config parameters ****
ARM_CHECK_UDF=true
GET_VIDEO_TITLE=true
SKIP_TRANSCODE=false
RAWPATH="/mnt/media/ARM/raw/"
MEDIA_DIR="/mnt/media/ARM/Media/Movies/"
LOGPATH="/mnt/media/ARM/logs/"
LOGLIFE=1
SET_MEDIA_PERMISSIONS=false
CHMOD_VALUE=777
MINLENGTH="600"
RIPMETHOD="backup"
MKV_ARGS="--profile=/opt/arm/default.mmcp.xml"
HB_PRESET="High Profile"
DEST_EXT=mkv
HANDBRAKE_CLI=HandBrakeCLI
MAINFEATURE=true
HB_ARGS="--subtitle scan -F -e x265 -q 20 -E copy:*"
PLEX_SUPPORT=false
EMBY_REFRESH=false
EMBY_SUBFOLDERS=true
EMBY_SERVER=""
EMBY_PORT="8096"
EMBY_CLIENT="ARM"
EMBY_DEVICE="ARM"
EMBY_DEVICEID="ARM"
EMBY_USERNAME=""
EMBY_USERID=""
IFTTT_EVENT="arm_event"
LOG=$LOGPATH$LOGFILE
*** End config parameters ****
Starting Identify Script...
Deleting 0 old log files: 
DEVLINKS=/dev/cdrom /dev/dvd /dev/disk/by-label/BEOWULFD1_NA /dev/disk/by-uuid/38322A3800000000 /dev/disk/by-id/ata-VMware_Virtual_SATA_CDRW_Drive_01000000000000000001 /dev/disk/by-path/pci-0000:02:04.0-ata-2
DEVNAME=/dev/sr0
DEVPATH=/devices/pci0000:00/0000:00:11.0/0000:02:04.0/ata4/host3/target3:0:0/3:0:0:0/block/sr0
DEVTYPE=disk
ID_ATA=1
ID_ATA_FEATURE_SET_PM=1
ID_ATA_FEATURE_SET_PM_ENABLED=1
ID_BUS=ata
ID_CDROM=1
ID_CDROM_CD=1
ID_CDROM_DVD=1
ID_CDROM_HDDVD=1
[...]

PR #60, opened 2017-05-19
Add HDDVD support to identify.sh

TheFiZi:
This has worked 3/3 at this point for me with HDDVDs.

I found in some cases the file system of the HDDVD is read-only and the standard -d check won't work on the HVDVD_TS directory which is why sometimes the ugly 'ls' command is required.

Resolves issue #59

### DIFF
--- a/identify.sh
+++ b/identify.sh
@@ -46,6 +46,7 @@ if [ "$ID_FS_TYPE" == "udf" ]; then
 		# check to see if this is really a video
 		mkdir -p /mnt/"$DEVNAME"
 		mount "$DEVNAME" /mnt/"$DEVNAME"
+		# shellcheck disable=SC2126
 		if [[ -d /mnt/${DEVNAME}/VIDEO_TS || -d /mnt/${DEVNAME}/BDMV || -d /mnt/${DEVNAME}/HVDVD_TS || $(ls -laR /mnt/${DEVNAME}/ 2>/dev/null |grep -P "HVDVD_TS" |wc -l) == 1 ]]; then
 			echo "identified udf as video" >> "$LOG"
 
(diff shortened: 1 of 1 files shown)


identify.sh
```diff
@@ -46,7 +46,7 @@ if [ "$ID_FS_TYPE" == "udf" ]; then
 		# check to see if this is really a video
 		mkdir -p /mnt/"$DEVNAME"
 		mount "$DEVNAME" /mnt/"$DEVNAME"
-		if [[ -d /mnt/${DEVNAME}/VIDEO_TS || -d /mnt/${DEVNAME}/BDMV ]]; then
+		if [[ -d /mnt/${DEVNAME}/VIDEO_TS || -d /mnt/${DEVNAME}/BDMV || -d /mnt/${DEVNAME}/HVDVD_TS || $(ls -laR /mnt/${DEVNAME}/ 2>/dev/null |grep -P "HVDVD_TS" |wc -l) == 1 ]]; then
```
ahnooie: This looks good, can you add "# shellcheck disable=SC2126" just above this line to disable the shellcheck warning?  Thanks.

TheFiZi:
Added the line as requested but I must have miss-understood. The check still failed.

ahnooie:
Nope, you fixed it.  It failed on a different check this time.  If you click on the "Details" link next to the build failure it will show why it failed...this time it failed on SC2010 because ls and grep won't work with non-alphanumeric filenames.  I don't think there would ever be a scenario here where we're dealing with non-alphanumeric characters here so if you want you can ignore that check by putting in another disable for SC2010 above that line.

ahnooie:
I didn't know people still had HDDVDs!  Thanks for the PR.

Issue #61, opened 2017-06-05
Minor (non-fatal) bug: Error reported to email (root@localhost) when non-audio CD is inserted


cbunt1:
When I insert a non-audio CD I get a report in root's email:
/opt/arm/config
/opt/arm/identify.sh: line 109: ((: > 0 : syntax error: operand expected
(error token is "> 0 ")

This is a non-fatal, very minor bug. It probably only bothers me because I'm a "clean log" purist. :)

Environment is (from uname -a)
Linux localhost 4.4.0-78-generic #99-Ubuntu SMP Thu Apr 27 15:29:09 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
Ubuntu server 16.04.1 LTS running on an older MSI laptop. Error reproduces on VirtualBox version of the same implementation.

I have looked through identify.sh and reworked line 109 and will submit a pull request for evaluation. The fix seems to eliminate the logged error while still properly identifying audio vs. non-audio discs.

PR #62, opened 2017-06-05
fix reported issue #61, non-audo-cd bug fix and add timestamp to logs

cbunt1:
Corrects error reported to root@localhost in log reported in issue #61 when a non-audio disc is inserted
Also added a timestamp to each logfile (line 20)


ahnooie:
Thanks, @cbunt1

PR #64, opened 2017-06-05
Feature audio cd identify

cbunt1:
Enhance the logging methodology for audio discs. If an audio disc is identified, the musicbrainz or cddb disc ID is used as a logging file name. 

This makes multiple rip processes easier to identify and troubleshoot, and puts audio ripping logs in position to be cleaned by the existing log-cleaning routine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ archive/
 config
 temp/
 test.py
+*.sh~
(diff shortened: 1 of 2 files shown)


ahnooie:
Thanks, this is useful.

Issue #57, opened 2017-04-11
Hancock ripped but did not get moved to "Movies"


MidasLamb:
The movie hancock got ripped and moved from raw to Media/Unidentified, but does not move to Media/Movies

cbunt1:
I have had this problem with every movie I've ripped, but I just set the appliance up a few days ago.

Upon investigation, it's because they're not being positively identified as a movie. In the log for each movie you'll see "video type is fail." As best I can tell, this is because the omdbapi.com site called by getvideotype.py has recently had to go private with their database and now requires an API key to complete a query. (My manual queries are denied due to lack of a key...so I'm concluding this is the reason.)

This appears to be new (since 2017-05-08) and for the time being, keys are available to patrons only ($1.00 per month) -- but the site host reports that they are planning to open it back up to non-paying customers for limited queries (the $1.00/month level buys 100,000 requests per day--far more than I'll use in a lifetime of ripping DVD's!).

MidasLamb:
Thanks for your response. The issue is only a minor inconvenience for this whole set-up. I only reported it so the author knows there's a bug. But thanks that you took the time to answer this :)

ahnooie:
Ah, thanks for reporting that.  Not sure if there's any way around it, I haven't found any other APIs that will work.

cbunt1:
I'm hoping that once they get everything settled down over at omdbapi that they'll open up the casual-use api again. I'm willing to spend a little money to enable it, if only for testing purposes, but the thresholds they're quoting are high enough to make me think that the situation is temporary. If they do, even a serious casual ripper who is converting his/her video collection wouldn't use but maybe 20-30 queries in a day.

I agree that its only a minor annoyance, but it had me scratching my head when I was trying to work out the details of setting one of these ARM boxes up. I'm not sure I can really contribute anything useful to the project, but I'll try where I can :)

ahnooie:
Thanks @cbunt1 , I'm a bit behind reviewing the PRs as it's summer and the days have been nice... but I'll try to take a look on the next rainy day.  I was thinking someone could put up a reverse proxy for the OMDB Api (just have the proxy limit each IP to 25 requests per day which should be plenty for most people to avoid going over the 100K limit) for ARM users to hit, but obviously my concern is what the OMDB API ToS allows, I couldn't seem to find their terms of use stated anywhere but I'm sure something like that is prohibited.  Long term themoviedb may be a better API since it has free API keys.

[...]

PR #65, opened 2017-06-09
Dep omdbapi key, Fixes root cause of issue #57.

cbunt1:
We have a fix for the root cause of Issue #57 . At this point I have purchased a year-long API key which I'm willing to share for testing (contact me directly). The fix follows the format of the other API keys and such by storing the key in the config file. Note that the PR version has been sanitized of the API key.


jswetzen:
I'd like to test this, will email you about the key.

cbunt1:
Sorry for my delayed response. The key is in the email, so to speak. :)

jswetzen:
I can confirm that it's working, and the code looks good to me as well.

cbunt1:
I think that should do it. Let me know if I removed files that we should have kept.

ahnooie:
Looks good to me, thanks for adding this!

jswetzen:
You lost requirements.txt, one text file too many.

cbunt1:
OOPS! 

I also pulled the README-OMDBAPI file. Let me know if I should re-add it, since it has the docs for using the api key.

ahnooie:
Oops, I missed the requirements as well.  I think it's a good idea to add the omdbapi readme back as well.

cbunt1:
I re-added requirements.txt and omdbapi-readme. They came up as a separate pull request, but I think I have everything fixed now.

PR #68, opened 2017-07-20
Avoid eject loop

jswetzen:
There was a problem with VirtualBox on Mac that caused the eject of a disk to
trigger the udev rule and script again. So it ejected, even though the drive was
empty, triggering the udev rule and script again...
I don't know why it appeared for me and not for anyone else, but I added an extra case in identify.sh to skip the eject in there doesn't seem to be a disk in the drive.


ahnooie:
Thanks!  Looks good.

PR #71, opened 2017-07-29
Dep omdbapi key

cbunt1:
Add README-OMDBAPI and requirements.txt back to bundle


ahnooie:
Thanks, for some reason it didn't bring requirements, I can see it in your changes so not sure why.  I'll do a separate PR for that.

PR #76, opened 2017-07-30
use systemd

kmarhold:
change udev rule to call systemd
add systemd service
get disc info in wrapper script
adjust identify and rip scripts
add systemd to installation instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ See: https://b3n.org/automatic-ripping-machine
     - Retrieve title from disc or Windows Media MetaServices API to name the folder "movie title (year)" so that Plex or Emby can pick it up
     - Determine if video is Movie or TV using OMDb API
     - Rip using MakeMKV (can rip all features or main feature)
-    - Eject disc and queue up Handbrake transcoding when done 
+    - Eject disc and queue up Handbrake transcoding when done
     - Transcoding jobs are asynchronusly batched from ripping
     - Send notification when done via IFTTT or Pushbullet
   - If audio (CD) - rip using abcde
@@ -58,12 +58,13 @@ If you have a  new DVD drive that you haven't used before, some require setting 
     pip3 install -r requirements.txt
     ln -s /opt/arm/51-automedia.rules /lib/udev/rules.d/
     ln -s /opt/arm/.abcde.conf /root/
+    cp /opt/arm/arm@.service /etc/systemd/system/
     cp config.sample config
 
 - Edit your "config" file to determine what options you'd like to use
 - To rip Blu-Rays after the MakeMKV trial is up you will need to purchase a license key or while MakeMKV is in BETA you can get a free key (which you will need to update from time to time) here:  https://www.makemkv.com/forum2/viewtopic.php?f=5&t=1053 and create /root/.MakeMKV/settings.conf with the contents:
 
-        app_Key = "insertlicensekeyhere" 
+        app_Key = "insertlicensekeyhere"
 
 
 Optionally if you want something more stable than master you can download the latest release from the releases page.
--- a/arm_wrapper.sh
+++ b/arm_wrapper.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 DEVNAME=$1
-udevadm info --query=env "/dev/$DEVNAME" > /tmp/arm_disc_info_$DEVNAME
-echo bash /opt/arm/identify.sh /opt/arm/config /tmp/arm_disc_info_$DEVNAME | at now
+udevadm info --query=env "/dev/$DEVNAME" > /tmp/arm_disc_info_"$DEVNAME"
+echo bash /opt/arm/identify.sh /opt/arm/config /tmp/arm_disc_info_"$DEVNAME" | at now
--- a/arm_wrapper.sh
+++ b/arm_wrapper.sh
@@ -1,2 +1,4 @@
 #!/bin/bash
-echo bash /opt/arm/identify.sh /opt/arm/config | at now
+DEVNAME=$1
+udevadm info --query=env "/dev/$DEVNAME" > /tmp/arm_disc_info
+echo bash /opt/arm/identify.sh /opt/arm/config /tmp/arm_disc_info | at now
--- a/arm_wrapper.sh
+++ b/arm_wrapper.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
-# shellcheck disable=SC1090
 DEVNAME=$1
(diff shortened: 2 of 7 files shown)


arm_wrapper.sh
```diff
@@ -1,2 +1,4 @@
 #!/bin/bash
-echo bash /opt/arm/identify.sh /opt/arm/config | at now
+DEVNAME=$1
+udevadm info --query=env "/dev/$DEVNAME" > /tmp/arm_disc_info
```
ahnooie: This has the potential to cause a race condition for people ripping with multiple optical drives ripping things in parallel.  Can you change the output to something like /tmp/arm_disc_info_$DEVNAME ?

README.md
```diff
@@ -17,7 +17,7 @@ See: https://b3n.org/automatic-ripping-machine
     - Retrieve title from disc or Windows Media MetaServices API to name the folder "movie title (year)" so that Plex or Emby can pick it up
     - Determine if video is Movie or TV using OMDb API
     - Rip using MakeMKV (can rip all features or main feature)
-    - Eject disc and queue up Handbrake transcoding when done
```
ahnooie: Can you clear out these extra whitespaces at the end of the lines? Here and a few other places.

kmarhold:
OK, finally fixed the shellcheck errors (next time I'll do it locally).

ahnooie:
@kmarhold where should I put the "arm@.service" file?

ahnooie:
And let me know any other steps you can think of...   I have no idea how to use systemd.

kmarhold:
The arm@service file should be copied into /etc/systemd/system. I also added that to the instructions in README.md. I don't know if symlinking works, I haven't tried it. 

Other than that nothing is needed: the @ makes the service a "template service" that must be called from elsewhere (udev in our case) with an argument, so it doesn't have to be enabled or anything. 

I figured it out following this post where someone did the same thing but with a backup script:

[](https://brokkr.net/2017/06/12/why-my-udev-backup-script-kept-failing-and-how-i-fixed-it/)

ahnooie:
Thanks for updating the readme @kmarhold   I did a checkout on your branch, followed your instructions, and started testing using systemd, I still need to test data discs, and audio, but so far it seems to work great with DVDs!

One thing we need to test is if it works when ripping from multiple DVD drives simultaneously.  I only have one drive so I can't test this feature, do you by chance have multiple drives?  If not I can see if anyone else here can test it.

One minor issue under /opt/arm/logs the log is named "empty.log" instead of "the movie title.log" from the ID_FS_LABEL variable.  I do see the ID_FS_LABEL var is set in the empty.log so not sure why the log file isn't being written with that name.

kmarhold:
@ahnooie good to hear that it works! I don't have multiple drives either, that's why I didn't think of the race condition in the first place. But having fixed that I can't think of a reason why it would fail. The systemd service could even be called simultaneously with two different instances.

You are right about the logs, I actually had that problem since I started using the arm. ID_FS_LABEL appears in the disk info file, too, so could that be a bug somewhere else?

kmarhold:
@ahnooie just added one minor change: we should be removing the disc info at the end of identify.sh (before I removed it from video_rip.sh, I had only been ripping DVDs and forgot about the data rip script).

ahnooie:
The id_fs_label issue seemed to be fixed for me after a reboot.  No testers with multiple-drives came forward but it passed all my tests and should work.  Thanks for the PR!

2long1:
I only recently discovered your project.  I have wanted something like this for years.  Thanks for the hard work.  I built a Ubuntu 16.04.03 LTS VM hosted on vSphere 6.5.  The host hardware is an HP Z600 into which I installed 2 BluRay drives.  So I am able to help with the testing of multiple drives.  I do not know if the bug is with Ubuntu or VMWare, but when I insert a disk in one drive, I get a change notification for both /dev/sr0 and /dev/sr1:

root@arm:~# udevadm monitor
monitor will print the received events for:
UDEV - the event which udev sends out after rule processing
KERNEL - the kernel uevent

KERNEL[102.386602] change   /devices/pci0000:00/0000:00:11.0/0000:02:01.0/ata4/host3/target3:0:0/3:0:0:0/block/sr1 (block)
KERNEL[102.387657] change   /devices/pci0000:00/0000:00:11.0/0000:02:01.0/ata3/host2/target2:0:0/2:0:0:0/block/sr0 (block)
UDEV  [103.257963] change   /devices/pci0000:00/0000:00:11.0/0000:02:01.0/ata4/host3/target3:0:0/3:0:0:0/block/sr1 (block)
UDEV  [103.259772] change   /devices/pci0000:00/0000:00:11.0/0000:02:01.0/ata3/host2/target2:0:0/2:0:0:0/block/sr0 (block)
KERNEL[103.430232] add      /module/crc_itu_t (module)
UDEV  [103.431141] add      /module/crc_itu_t (module)
KERNEL[103.445836] add      /module/udf (module)
KERNEL[103.445984] add      /kernel/slab/udf_inode_cache (slab)
UDEV  [103.446669] add      /module/udf (module)
UDEV  [103.446690] add      /kernel/slab/udf_inode_cache (slab)
KERNEL[103.525242] add      /module/nls_utf8 (module)
UDEV  [103.526188] add      /module/nls_utf8 (module)
KERNEL[139.340550] change   /devices/pci0000:00/0000:00:11.0/0000:02:01.0/ata4/host3/target3:0:0/3:0:0:0/block/sr1 (block)
UDEV  [139.570783] change   /devices/pci0000:00/0000:00:11.0/0000:02:01.0/ata4/host3/target3:0:0/3:0:0:0/block/sr1 (block)
KERNEL[150.864625] change   /devices/pci0000:00/0000:00:11.0/0000:02:01.0/ata3/host2/target2:0:0/2:0:0:0/block/sr0 (block)
UDEV  [151.158000] change   /devices/pci0000:00/0000:00:11.0/0000:02:01.0/ata3/host2/target2:0:0/2:0:0:0/block/sr0 (block)

The dock panel then displays multiple dvd's with the same name.  I am going to test this on a linuxmint install and a Rhel 7.4, in hopes of determining why it is behaving this way.  Just thought I would let you know I am testing this now.

audioeng:
Odd, did you somehow configure the VM with the same host drive twice
instead of each host drive? That is what the description sounds like to me.

I am running on bare metal with a SATA Blu-ray and USB DVD drive and they
both work correctly (sr0 and sr1)

2long1:
Good catch, I thought I double checked it when configuring it with the web client, but apparently not.  I am retesting now.

PR #84, opened 2017-08-31
Add second notification point on rip complete

TheNuttyP:
Added a call to the notification script to run on rip completion so that the user is notified both when the rip AND the transcoding is completed.


ahnooie:
I agree with making this notification an option in the config.  I think the two use cases are a user wants to know when the rip is complete so they can insert the next disc (this PR) or a user wants to know when transcoding is done so they can watch the movie.

audioeng:
Should be a simple if check, I’ll look at putting together a PR for this, I
might have time tomorrow.

I agree with those two use cases, those are the two I like to have at
least. :)

audioeng:
Just submitted #88 PR to make these notifications configurable in the config file. Also added a rip notification to data_rip.sh for consistency.

Only notification that is not around for ripping is the audio rip, as it is not in a separate script. It could be added, but thought some discussion might be warranted for that as a separate PR.

ahnooie:
Thanks!  Closing in favor of #88.

PR #86, opened 2017-09-18
Change LABEL variable in LOG and Notify message at transcode completion.

### LINKED ISSUE #85: Log / Notification Doesn't Include Title Name
While working with ARM today, I found that the log only contains the title for the MakeMKV portion of the job, and not for the HandBrakeCLI/transcode portion.

The variable `${ID_FS_LABEL}` is used to contain the title name and is used in several scripts - `video_rip.sh`, `video_transcode.sh` - and while it's usage in `video_rip.sh` does result in the title being included in the log message, it doesn't when used in `video_transcode.sh` - I'm not sure if this is because the variable is becoming unset between one script and the next or what - I'm rather new to shell scripting.

It's not a breaking inconsistency, just triggers a bit of OCD :-)

audioeng:
Fixes #85 

ID_FS_LABEL is sometimes blank and not used anywhere else in this script, but the LABEL variable is used extensively in this script and will always have a value that matches the directory of the transcoded files.

This has brought my Transcode notification naming consistency to 100% from much lower as shown in these photos.

img_006968144ed9-1

img_638f969616e8-1

### DIFF
--- a/video_transcode.sh
+++ b/video_transcode.sh
@@ -295,8 +295,8 @@ TRANSEND=$(date +%s);
 TRANSSEC=$((TRANSEND-TRANSSTART));
 TRANSTIME="$((TRANSSEC / 3600)) hours, $(((TRANSSEC / 60) % 60)) minutes and $((TRANSSEC % 60)) seconds."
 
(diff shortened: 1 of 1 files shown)


ahnooie:
Thanks for the PR!

PR #88, opened 2017-09-22
Added Notification Confiuration

audioeng:
Added NOTIFY_RIP variable for post rip of video and data and NOTIFY_TRANSCODE variable for post transcode of videos. Default in the config.sample file is true for both values.

Tested this out and is working for me. Let me know if anything doesn't meet your naming/formatting requirements and I will happily change it.


ahnooie:
Thanks!

Issue #92, opened 2017-10-05
Notify when done ripping to FLAC?


CarterBrehm:
This software is really great, but I have a lot of discs I need to convert and I'd like to have some system of notification when it's done. Each disc takes a different amount of time and I basically need to babysit it for  a couple hundred discs. Is there a file I can modify to notify Pushbullet when a rip is done?

audioeng:
@carterbrehm 
Yes. I have to get the kids to bed, but I should be able to tell you what to add and where in the next few hours. I asked about it in #84, I can easily take what I did for #88 and make a new PR to have a notify at end of audio rip.

audioeng:
@CarterBrehm Please check out the PR I just submitted for this and compare, you will see the 3 lines needed in identify.sh to add this in from current master.

audioeng:
Example of what the notification looks like in Pushover. Will be similar in whatever notification tool you are using.
![img_0260](https://user-images.githubusercontent.com/17124213/31261956-badba0f6-aa25-11e7-9bdc-4a268c1120e2.jpg)

audioeng:
This has been added in PR #93 that has now been merged, think this is safe to close.

PR #93, opened 2017-10-06
Notify on completion of audio rip

audioeng:
This checks if you have rip notifications enabled and sends a notification after abcde is done making FLAC files, addresses #92


ahnooie:
Looks good, thanks.

PR #95, opened 2017-10-12
added email notification options

JSylvia007:
Ben,

I think I'm a quick student... First go with Git...  I Forked your repo over to my Git, created a branch, made the changes to notify.sh and config.sample.  I tested these out on my system, and they work.

Hopefully this get over to you for incorporation.

### DIFF
--- a/51-automedia.rules
+++ b/51-automedia.rules
@@ -1,5 +1,4 @@
 # ID_CDROM_MEDIA_BD = Bluray
 # ID_CDROM_MEDIA_DVD = DVD
 # ID_CDROM_MEDIA_CD = CD
-ACTION=="change", SUBSYSTEM=="block", RUN+="/opt/arm/arm_wrapper.sh"
-
+ACTION=="change", SUBSYSTEM=="block", TAG+="systemd", RUN+="/bin/systemctl start arm@%k.service"
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,25 @@
+# Contributing Guide
+## Introduction
+Thank you for contributing to the Automatic Ripping Machine.
+
+## Issues, Bugs, and Feature Requests
+If you find a bug, please paste the relevant logfile in /opt/arm/logs/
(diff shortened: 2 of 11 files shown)


audioeng:
I like the idea, but I am curious as to the need for mutt. Seems like this could be done with only postfix.

JSylvia007:
Not sure how you could do this with just postfix as postfix is only the mail server.  It has no way to send mail itself, it's merely the transport service.  

I've toyed around with a bunch of different command line mailers, and mutt seems to handle everything I've thrown at it, including complex HTML/XML and attachments.

audioeng:
Here is a good example of configuring postfix to [send an email from a gmail account.](https://easyengine.io/tutorials/linux/ubuntu-postfix-gmail-smtp/)

JSylvia007:
Right...  Notice that the command for testing echo's something to "mail"?  You still need a client. I prefer mutt over mail, that's all.

audioeng:
Ahh, but mail is already there, so you don't have to install and configure
something else. If you are sending via a command line script, whether you
like the mail client better is irrelevant, you don't see it. Ease of
configuration should be primary I think.

JSylvia007:
Technically, mail is added as part of mailutils, also referenced in that guide.  As I mentioned previously, mutt handles richly formatted emails and attachments, as well as more complex workflows far more elegantly than mail does.  Additionally, no configration is required, it's just simply `apt-get install mutt`

JSylvia007:
Awesome... I think you'll like mutt for flexibility.  The reason that I use postfix is because I have multiple things setup in a 10's of servers environment and it just makes it easier.

I just made and pushed those changes.

ahnooie:
Can you add a link to a mutt guide (the one @audioeng linked to is fine) in the config.sample file so people who use this option will know how to configure mutt?  Once that's done it looks good to me.

JSylvia007:
Language added!

JSylvia007:
Thanks @ahnooie !!

I've loved contributing to this project!!

PR #96, opened 2017-10-12
modified the 'at' command to be silent

JSylvia007:
Added the "-M" switch to the "at" command to stop the email outputs.

### DIFF
--- a/arm_wrapper.sh
+++ b/arm_wrapper.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 DEVNAME=$1
 udevadm info --query=env "/dev/$DEVNAME" > /tmp/arm_disc_info_"$DEVNAME"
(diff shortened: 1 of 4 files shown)


audioeng:
Huh, since this runs as root I never thought to check sudo mail, turns out that my mailbox is up to 700MB so far in less than 1000 messages, so this is probably a good change to make.  

I am in favor of a merge of this, I might need to add this to my PR #93 which adds a notification on audio rip as well.

JSylvia007:
Yea... That's what was scaring me... I keep a small home partition for security and standardization reasons, and I was looking at that getting filled quite quickly.

This is really my first foray into GitHub.  I think I may expose some of my other projects here.

audioeng:
Welcome to Github! I've got a few projects myself but this is the first one I have contributed to as well.

ahnooie:
@audioeng @JSylvia007 That's why I always provision my VMs with 100GB hard drives. |:-)  Thanks for mentioning this project got you into contributing on GitHub--that's music to my ears.

PR #98, opened 2017-10-15
Move logging logic out of config file

### LINKED ISSUE #89: Logic inside config file
There is an if statement at the end of the config file, what is the purpose of that and can it be moved elsewhere? Generally you wouldn't want a config file to contain business logic, only variables and values.

In trying to parse the config file using Python everything I have tried is choking on this if logic.
I don't know enough about bash scripting to know, but I know that all I have ever seen is an empty.log file so I am wondering if this is running for anyone?

```
# Determine logfile name
# use the label of the DVD/CD or else use empty.log
# this is required for udev events where there is no media available
# such as ejecting the drive

if [ -n "$ID_FS_LABEL" ]; then
        LOGFILE=${ID_FS_LABEL}".log"
        elif [[ -n "$ID_CDROM_MEDIA_TRACK_COUNT_AUDIO" && $(which abcde-musicbrainz-tool) ]]; then
                LOGFILE=$(abcde-musicbrainz-tool --device "$DEVNAME" | cut -f1 -d ' ')".log"
        elif [[ -n "$ID_CDROM_MEDIA_TRACK_COUNT_AUDIO" &&  $(which cd-discid) ]]; then
                LOGFILE=$(cd-discid "$DEVNAME" | cut -f1 -d ' ')".log"
else
        LOGFILE="empty.log"
fi

# Set full logfile path
LOG=$LOGPATH$LOGFILE
```

audioeng:
Fixes #89 and as a by product, gets the per disc logging (the purpose of the logic) working correctly! It accomplishes this by moving the logic to the identify.sh script. Whenever this calls another script, it passes the LOG variable to that script. I have edited each of these scripts to use this parameter, so all of a discs activity will be logged to its own file.

Adds a new variable to config.sample as well, `LOG_SINGLE_FILE`, which defaults to false. With false, you get per disc logging and empty disc goes to empty.log
If you set it to true, it will log **everything** to ARM.log, much like the behavior before with everything going to empty.log

I thought some people might like the option to keep an eye on a single log file vs individual and this was an easy way to offer that choice.

Lots of changes for this to several files, I have double checked as well as tested out all the different disc types and it is working in my tests. Please test as well!

### DIFF
--- a/config.sample
+++ b/config.sample
@@ -39,6 +39,9 @@ LOGPATH="/opt/arm/logs/"
 # How long to let log files live before deleting (in days)
 LOGLIFE=1
 
+# Set to true if you prefer a single log file for all activity versus a separate log per disc.
+LOG_SINGLE_FILE=false 
+
 ########################
 ##  File Permissions  ##
 ########################
--- a/config.sample
+++ b/config.sample
@@ -175,21 +175,5 @@ PO_APP_KEY=""
 OMDB_API_KEY=""
 
 
-# Determine logfile name
-# use the label of the DVD/CD or else use empty.log
-# this is required for udev events where there is no media available
-# such as ejecting the drive
 
-if [ -n "$ID_FS_LABEL" ]; then
-        LOGFILE=${ID_FS_LABEL}".log"
-        elif [[ -n "$ID_CDROM_MEDIA_TRACK_COUNT_AUDIO" && $(which abcde-musicbrainz-tool) ]]; then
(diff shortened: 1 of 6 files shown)


JSylvia007:
This looks really great.  I love the idea of getting the logic out of the config!

audioeng:
Has anyone else had a chance to test this PR yet? It has been working well for me but would love any feedback you may have.

ahnooie:
Thanks!

audioeng:
My pleasure, love this project!

Issue #101, opened 2017-11-12
Udev rules causing systemd-udev-settle.service to timeout


nsteffan:
Hello, I am running Ubuntu 17.04. I noticed after installation that the systemd-udev-settle.service was timing out as my ZFS mounts weren't coming up. It appears that the udev rules aren't specific enough and are passing all block change events to the arm service, instead of just those related to the CD drives. I looked at the cdrom_id udev rules and noticed we can add some specificity. I fixed it by changing the udev rules to this:

`ACTION=="change", SUBSYSTEM=="block", TAG+="systemd", KERNEL=="sr[0-9]*|vdisk*|xvd*", ENV{DEVTYPE}=="disk", RUN+="/bin/systemctl start arm@%k.service"`

I can submit a PR if needed.

ahnooie:
Nice!  A PR would be great.

PR #103, opened 2017-11-21
Fix udev rules blocking udev settle service

nsteffan:
Fixes issue #101 by specifying the kernel target and devtype disk target.


ahnooie:
Looks good, thanks!

PR #114, opened 2018-02-01
readme updates

ahnooie:
A few changes to the readme after my first test install on a blank system.  May find more so don't merge it in yet.

### DIFF
--- a/docs/config.sample
+++ b/docs/config.sample
@@ -87,7 +87,6 @@ RIPMETHOD="mkv"
 # audio and burn them into the video
 # HB_ARGS="--subtitle scan -F --subtitle-burned"
 # HB_ARGS= "--subtitle scan -F --audio-lang-list eng --all-audio"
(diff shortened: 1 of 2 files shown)


ahnooie:
I'll probably have to take a 2nd pass from a clean slate to make sure but this PR is ready... we may need to do some more to the config file sample but I know I had to make these changes.

docs/config.sample
```diff
@@ -87,7 +87,6 @@ RIPMETHOD="mkv"
 # audio and burn them into the video
 # HB_ARGS="--subtitle scan -F --subtitle-burned"
 # HB_ARGS= "--subtitle scan -F --audio-lang-list eng --all-audio"
-HB_ARGS="--subtitle scan -F --subtitle-burned"
```
ahnooie: Should this be HB_ARGS_DVD?

PR #120, opened 2018-02-06
V2bugs

muckngrind4:


audioeng:
I like where this is going, moving to python. I did notice that the code that I had removed from the config file to fix the log files not generating correctly is back in this config. Planning on moving this back out? Logic really shouldn't live inside a config file, and especially this code which deals with setting variables based on which disc has been inserted. It really belonged in the identify code.

muckngrind4:
@audioeng Funny you should mention it. I'm working on that right now.  I've been using an old version of the config file (stupid me!) and am now updating it to be consistent with new versions and removing anything that's no longer needed, such such log file generations.  That logic is now where is should be--in the python code.   

BTW, if you're bored and somewhat adventurous, the v2_master branch is mostly complete and going into a testing phase.  That branch will become V2.0.0.  The goal is to release it in late April around the same time as Ubuntu 18.04.  Feel free to grab it, play with it, give feedback, commit, whatever you want!  It's missing a couple of things and I want to clean up some code, but it should be working enough to test. ;)

audioeng:
I've noticed the flurry of activity, I'm not bored but definitely up for building a new VM with this v2 branch to help test!

Recommend using 16.04 or 17.10 at this point?

muckngrind4:
I'm running 16.04 now, but will soon build a 17.10 vm to get a feel for how it will run on 18.04.  So, if you're willing, 17.10 would be great.  The thing to watch for is how udev and systemd work.  We've reduced the need for udev, however it still requires a udev rule to launch systemd, so I'm not sure how that will work with all the udev changes in Ubuntu.  Submit any issues and I'll try to help direct you where to look.  Thanks!

PR #127, opened 2018-02-08
Added a check for existing file before moving to avoid overwrite.

### LINKED ISSUE #121: V2 overwrites existing file in media directory
Need a check to make sure the file doesn't exist before moving it.

muckngrind4:
Closes #121


muckngrind4:
@ahnooie Check out the first failure above.  Is this an issue in code somewhere or is it in the config of Travis?

PR #130, opened 2018-02-08
Issue template 1

ahnooie:
@muckngrind4 I think we need an issue template, at least to prompt for logs and config files.  What do you think of something like this?


muckngrind4:
Totally agree.  We should probably include a reminder to sanitize the config file before attaching.

PR #159, opened 2018-04-25
Fix #104 by exporting disc info variables

### LINKED ISSUE #104: ARM broke after upgrades
I'm not 100% where the issue lies because like an idiot I upgraded ARM and MakeMVK at the same time with out testing in between.

Previous to the upgrade discs were ripping fine.

I've attached the log. I'm not 100% certain if the issue is HandbreakCLI, MakeMKV or ARM itself.

ARM is seeing a disc has been inserted and is calling MakeMKV. MakeMKV appears to rip the disc successfully. The failure occurs at the transcoding portion.

I've re-compiled my HandbreakCLI just in case but that hasn't helped.

Log attached: 
[failedRip.log](https://github.com/ahnooie/automatic-ripping-machine/files/1497817/failedRip.log)

I'm thinking the issue is this:

```
[20:11:26] hb_scan: path=/mnt/ripTemp/raw//LOGICAL_VOLUME_ID_20171121_192805/CERTIFICATE, title_index=1
disc.c:352: error opening file BDMV/index.bdmv
disc.c:352: error opening file BDMV/BACKUP/index.bdmv
```

Some quick digging turned up that this might be a Handbreak issue?

jswetzen:
This fixes #104 because it exports the necessary variables so that they carry over to video_transcode.sh.

### DIFF
--- a/video_rip.sh
+++ b/video_rip.sh
@@ -4,8 +4,8 @@
 # shellcheck source=config
 # shellcheck disable=SC1091
 source "$ARM_CONFIG"
-# shellcheck disable=SC1090
-while read info_var; do export $info_var; done < "${DISC_INFO}"
+# shellcheck disable=SC2163
+while read -r info_var; do export $info_var; done < "${DISC_INFO}"
 
 VIDEO_TITLE=$1
 HAS_NICE_TITLE=$2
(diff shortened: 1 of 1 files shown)


ahnooie:
LGTM, thanks!

Issue #113, opened 2018-01-30
Stop disc being re-read


martin8777:
Is there a way to stop the DVD from automatically closing itself? 

I'm finding that discs are being ripped multiple times if I don't go and take it out quickly enough. It seems like the disc drawer is automatically closing after a period of time and then the script just rips it and queues it for transcoding again. 

Is there anyone to prevent the DVD drive auto closing? and is it possible to tell ARM not to rip a disc it already has a .log file for ?

thanks.

muckngrind4:
I think some drives automatically pull the drive back after a certain timeout.  I suspect this is what is happening.  Making the eject an option in the config file is probably a good idea or adding logic to check if the inserted dvd was just ripped successfully.  I'll mark this as a enhancement request.  In the short term, you could try commenting out the eject lines in the scripts.  Put a # in front of any line with the "eject "$DEVNAME" command in the video_rip, video_transcode and identify files.  That should keep ARM from automatically ejecting after it's complete.

martin8777:
Yeah, I was thinking of commenting out the eject command since I'm getting notified of when it's done anyway but it is quite nice to be able to tell my wife "if you see a disc ejected, can you take it out and pop in the next one from the big pile!"

audioeng:
Hmm, are you using Ubuntu? What version? Physical or virtual? Any bios
settings regarding disk behavior?

martin8777:
Ubuntu 16.04 LTS physical install. I'll check out the bios again later but pretty sure there's no options like that in there. It's an old HP 6000 SFF PC.

muckngrind4:
@martin8777 Once it ejects, how long transpires before it's pulled back in?

audioeng:
I’m running an old dell and an LG Blu-ray and it stays ejected for days at
least...

martin8777:
looks like it auto closes after 10 minutes

didn't get a chance to check bios last night. grown up stuff took over,
will try tonight.

muckngrind4:
I think the proper way to handle this is with some logic that identifies DVD's that have already been ripped.  I'd like to implement something like that in the not to distant future...

martin8777:
Yeah, that would be nice. If I had any coding skill myself, I'd help out, but I'm more the Ops side of DevOps!

[...]

PR #193, opened 2018-07-03
Added experimental Auto grabbing of MakeMKV beta keys + double rip prevention.

duplexsystem:
MakeMKV auto grabbing Still needs some work:
- Option to enable / disable needs to be moved to config file.
- Might mess up during times when there are two rips taking place at the same time (need to add fail safe)
- Overwrites MakeMKV config file in stead of changing it
- If key grab fails will still overwrite MakeMKV config file with blank of blotched key.
Fixed double ripping #113


muckngrind4:
Hi @budgidiere, I just created a v2.1_dev branch and I'll likely merge this into that branch tomorrow.  I'm trying to stabalize v2_master for a release.  Thanks for the contribution!

aron7676:
Suggestion, make the pickle data a tuple, with (discdrive, str(disc.videotitle)) as the data. 
ex (sr0, avatar)
on script start, check if pickle[0] = disc and pickle[1] = title. 
Should fix the issue, just dont have the time to do the fork, update/test code/ push.

1337-server:
No longer needed as MakeMKV already handles this.

duplexsystem:
#192 should be reopened then

1337-server:
Part of #113 is taken care of with this #294 (Stop Drive door from auto closing)


The rest such as dupe rip protection are here
-------------------
This version kills instant duplicate rips - When the udev triggers twice quickly (<1 mins on the same devpath (/dev/sr0))
https://github.com/automatic-ripping-machine/automatic-ripping-machine/blob/cc52f90e991047ddd6ab4e1b024d2ebf795ae063/arm/ripper/main.py#L405


This version stops jobs that are already in the transcoding path
-------------------
https://github.com/automatic-ripping-machine/automatic-ripping-machine/blob/cc52f90e991047ddd6ab4e1b024d2ebf795ae063/arm/ripper/main.py#L217



But in the more updated v2_devel branch of ARM, there is a more complicated version of the have dupes check here
-------------------
https://github.com/automatic-ripping-machine/automatic-ripping-machine/blob/158dc70a01635c63c062738877be8a03470eff6c/arm/ripper/main.py#L179
which triggers this
https://github.com/automatic-ripping-machine/automatic-ripping-machine/blob/158dc70a01635c63c062738877be8a03470eff6c/arm/ripper/utils.py#L543


I'm sure this will cover most cases, and adding more checks on top of these already excessive check is unnecessary I think.

Issue #202, opened 2018-07-13
Duplicate kernel events when inserting disc causing ARM to run twice


jaffee:
### Description
I got ARM working after a bit of difficulty, was able to rip a few discs, then they stopped working properly. Long story short, the kernel seems to be firing two events when I insert a DVD which is causing two ARM jobs to get fired for the same device which causes a variety of havoc.

Here is udevadm monitor showing the duplicated events.
```
# udevadm monitor
monitor will print the received events for:
UDEV - the event which udev sends out after rule processing
KERNEL - the kernel uevent

KERNEL[18803.960118] change   /devices/pci0000:00/0000:00:01.1/ata1/host0/target0:0:0/0:0:0:0/block/sr0 (block)
KERNEL[18804.070848] change   /devices/pci0000:00/0000:00:01.1/ata1/host0/target0:0:0/0:0:0:0/block/sr0 (block)
UDEV  [18805.041536] change   /devices/pci0000:00/0000:00:01.1/ata1/host0/target0:0:0/0:0:0:0/block/sr0 (block)
UDEV  [18805.690674] change   /devices/pci0000:00/0000:00:01.1/ata1/host0/target0:0:0/0:0:0:0/block/sr0 (block)
```

I'm running Ubuntu 16.04 in VirtualBox on Windows Server 2016 Essentials.

I realize this probably isn't directly related to ARM, but I think ARM can probably work around it and I may try to submit a patch. What's odd to me is that the problem occurred several times, then stopped for a few discs, and now is occurring again.

### Log file
Here is a snippet - you can see how lots of lines are duplicated:

```
identified udf
found O200NNW1 on /dev/sr0
identified udf as video
identified udf as video
Obtained Title Oliver and Company (2009)
Obtained Title Oliver and Company (2009)
got to here
HAS_NICE_TITLE is true
video title is now Oliver and Company (2009)
video type is movie
got to here
HAS_NICE_TITLE is true
video title is now Oliver and Company (2009)
video type is movie
Starting video_rip.sh
Video Title is Oliver and Company (2009)
Ripping video O200NNW1 from /dev/sr0
Using mkv method of ripping.
Starting video_rip.sh
Video Title is Oliver and Company (2009)
Ripping video O200NNW1 from /dev/sr0
Using mkv method of ripping.
MSG:1005,0,1,"MakeMKV v1.12.2 linux(x64-release) started","%1 started","MakeMKV v1.12.2 linux(x64-release)"
[...]

PR #215, opened 2018-07-18
202 dup event protection

jaffee:
for posterity at least, here is what I ended up doing for #202 

It seems to work, but I ran into a number of other issues running Ubuntu inside virtualbox with multiple optical drives on Windows 10. Windows kept freezing and I had to reboot the machine. Ultimately I scrapped it, installed Ubuntu 16.04 directly on the host, and installed the v2_master branch of ARM which seems to be working.

I'm not sure I'd even recommend merging this as it is possible for it to get into a bad state where the lock directory is not deleted (though it tries to have a helpful message in this case). Some more complex logic that waited a while if the lock couldn't be acquired and then checked to see if an instance of makemkv was already running for the drive in question could make this nicer, but it was more complexity than I wanted to deal with.

### DIFF
--- a/arm_wrapper.sh
+++ b/arm_wrapper.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 DEVNAME=$1
 udevadm info --query=env "/dev/$DEVNAME" > /tmp/arm_disc_info_"$DEVNAME"
-echo bash /opt/arm/identify.sh /opt/arm/config /tmp/arm_disc_info_"$DEVNAME" | at -M now
+echo bash /opt/arm/identify.sh /opt/arm/config /tmp/arm_disc_info_"$DEVNAME" "$DEVNAME" | at -M now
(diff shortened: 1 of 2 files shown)


Plexen85:
Maybe installing Linux as a subsystem to Windows 10 will work better for you then virtual box? I haven’t done it yet, but is on my list of things to try when a weekend frees up.

muckngrind4:
Thanks @jaffee.  Glad you're working, even if it is on 16.04 native.  I agree, we'll probably leave it as a PR unless/until the Windows/virtualbox environment can find stability.

cmbernard333:
I ran into a similar issue with this where I put a bluray into my drive and it spun up three different processes trying to read /dev/sr0. I had to manually kill processes until I got down to just one. This would be a welcome change.

1337-server:
> I ran into a similar issue with this where I put a bluray into my drive and it spun up three different processes trying to read /dev/sr0. I had to manually kill processes until I got down to just one. This would be a welcome change.

pull #373 has a similar function, but in python. It allows preventing of dupe rips, which when enabled would stop all further processes from continuing. You can enable and disable it easily via the ARM config yaml.

PR #248, opened 2018-09-09
identifies the title of a music cd and sets the label of the disc-object

imthemichi:
similar to getvideotitle.py I have added getmusictitle.py, which is called from identify.py. In identify.py the label attribute of the disc object is set. This label is used during notification and now display to the user.


imthemichi:
mh on ubuntu I had to install libdiscid-dev

1337-server:
This PR is already added

PR #251, opened 2018-09-29
Fix typo when logging that data disc was found.

Wingysam:


muckngrind4:
Thanks for the PR!  I've implemented the change in the dev branch.

PR #258, opened 2018-10-15
Update README.md to add "sudo" to one line

amlamarra:
Sudo is needed for the command `cp docs/arm.yaml.sample arm.yaml`


flammableliquids:
Thank you.
Incorporated and no longer required.

PR #314, opened 2019-11-13
framework and logging reorganization

accountnumber3:
Pretty sure `scripts/handbrake_isoTOmp4.sh` doesn't actually work.
made EJECT optional
`scripts/manual.sh` probably doesn't work either
`scripts/update_makemkv.sh` automate license update. Change browser agent string?
`setup/.abcde.conf` I don't know why this changed. I don't rip music
`setup/install.sh` simplify setup and make it easier to reference file locations

### DIFF
--- a/arm/classes.py
+++ b/arm/classes.py
@@ -65,9 +65,11 @@ class Disc(object):
         return s
 
     def eject(self):
-        """Eject disc if it hasn't previously been ejected"""
-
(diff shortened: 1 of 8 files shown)


accountnumber3:
I didn't actually mean to submit this, I was trying to push to my fork and I just suck at git.

PR #318, opened 2019-12-29
Implement an override setting for the identified disc type

maluramichael:
I want to rip alot of cds with games on them from magazines. ARM identifies many of them as music dics and fails. This is why i implemented an override for the identification.


flammableliquids:
I think this should be built into the manual override function of the gui.

flammableliquids:
This is being dealt with in in 373.

PR #319, opened 2019-12-29
Replace cat command with dd for more control

maluramichael:
I need a bit more control over the copy process so i replaced cat with dd. This way i can add a `conv=noerror,sync` in which case the copy process is not interrupted. See [https://superuser.com/a/622553](https://superuser.com/a/622553).


flammableliquids:
This can no longer be merged into the master as utils.py has moved. 
I have added it to my repo and it should come with my pull request.

PR #320, opened 2019-12-29
Unlink iso after rip_data failed

maluramichael:


1337-server:
This PR is already added

PR #337, opened 2020-04-27
Fixed error with werkzeug lib.

dijitalmunky:
This is temporary, but it will at least allow users to work out of the box.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,5 +17,4 @@ omdb>=0.10.0
 psutil>=5.4.6
 robobrowser>=0.5.3
 tinydownload>=0.1.0
-
-
+werkzeug==0.16.0


kiwinv:
I ran into this issue on Ubuntu 18.04 and fixed it by adding the following lines to main.py

> import werkzeug
> werkzeug.cached_property = werkzeug.utils.cached_property

I'm not a developer so I'm not sure whether this is a viable long-term fix or not but it worked for me

dijitalmunky:
I am a developer, but not so much with python (I am a Java/C/C#/Javascript kind of guy).  However, I would say this is also a decent workaround until the libraries that depend upon werkzeug get fixed and this project upgrades them.

The only thing risked that I can see by doing this is that some other incompatible change was made in werkzeug that causes other issues.

j4nr6n:
Related: #347

flammableliquids:
no longer required in v2. [refer-here](https://github.com/flammableliquids/automatic-ripping-machine/pull/1) Tested out fine on 20.04
please get latest and check agian.

PR #338, opened 2020-05-09
Added logic for trying to parse raw DVD title using OMDB instead of windowsmedia services because it is dead.

dijitalmunky:
Modified omdb api logic in case there is no year to start.

Added 16X9 to title things to strip.

Handled a rare condition where a DVD does not actually have a name/label.

Stop OMDB lookups when we cannot get a proper title from the DVD Label

Added a logging error when a mainfeature is not found.


emmakat:
Hi there, just curious if anyone has a working copy of this project? #342 and #343  are blockers, and I can't proceed. No idea how to find/fix all these errors. Anyone?

flammableliquids:
Due to the way the cookie has crumbled this change will be both rejected and incorporated into the master.
I incorporated what I could.
Thank you for your code fixes.

Issue #348, opened 2020-07-02
Using mkv mode and skipping transcode errors when moving mkv


rajlaud:
### Description
There is a bug in the logic to move the finished file when using `mkv` mode and skipping transcode. The wrong number of parameters is passed to `utils.move_files`. Will submit PR with fix.

### Environment
Debian Linux. arm v2.1_dev with pull request #338 .

### Log file
[2020-07-01 15:22:01] INFO ARM: Exiting MakeMKV processing with return value of: /home/arm/media/raw/The Lego
Movie
[2020-07-01 15:22:01] INFO ARM: SKIP_TRANSCODE is true.  Moving raw mkv files.                               
[2020-07-01 15:22:01] INFO ARM: NOTE: Identified main feature may not be actual main feature                 
[2020-07-01 15:22:01] ERROR ARM: A fatal error has occured and ARM is exiting.  See traceback below for detail
s.
Traceback (most recent call last):
  File "/opt/arm/arm/ripper/main.py", line 416, in <module>                                                  
    main(logfile, job)
  File "/opt/arm/arm/ripper/main.py", line 207, in main                                                      
    utils.move_files(hbinpath, f, job.hasnicetitle, job, True)                                               
TypeError: move_files() takes from 3 to 4 positional arguments but 5 were given

[>>>>> DRAG AND DROP LOG FILE HERE <<<<<]

PR #349, opened 2020-07-02
Pass correct number of arguments to utils.move_files

rajlaud:
Fixes error encountered when skipping transcode and using `mkv` mode. Fixes issue #348


flammableliquids:
Due to the way the cookie has crumbled this change will be both rejected and incorporated into the master.

PR #352, opened 2020-07-11
Update to arm_wrapper.sh

flammableliquids:
This modification starts the Webui if it isn't already started.
Fixing a frequently occurring annoyance.


flammableliquids:
superseeded

PR #365, opened 2020-11-09
run ARM in Docker

deekue:
any interest in including the Docker bits?

### DIFF
--- a/arm/ripper/main.py
+++ b/arm/ripper/main.py
@@ -26,6 +26,8 @@ def entry():
     """ Entry to program, parses arguments"""
     parser = argparse.ArgumentParser(description='Process disc using ARM')
     parser.add_argument('-d', '--devpath', help='Devpath', required=True)
+    parser.add_argument('-t', '--disctype', help='udev disctype', required=False)
+    parser.add_argument('-l', '--label', help='udev label', required=False)
 
     return parser.parse_args()
 
@@ -378,6 +380,17 @@ if __name__ == "__main__":
     print(devpath)
 
     job = Job(devpath)
+
+    # override disctype from commandline (required in Docker, no udev)
+    udev = {}
+    if args.disctype:
+        (k, v) = args.disctype.split('=', 1)
+        udev[k] = v
+    if args.label:
+        (k, v) = args.label.split('=', 1)
+        udev[k] = v
+    if args.disctype or args.label:
+        job.set_disctype(udev.items())
 
     logfile = logger.setuplogging(job)
     print("Log: " + logfile)
--- a/scripts/docker-entrypoint.sh
+++ b/scripts/docker-entrypoint.sh
@@ -1,0 +1,50 @@
+#!/bin/bash -x
+#
+# set up a user and switch to that user
+
+set -euo pipefail
+
+UID="${UID:-1000}"
+GID="${GID:-1000}"
+export USER=arm
+export HOME="/home/${USER}"
+
+echo "creating group [arm] with id ${GID}"
+groupadd -fo -g "${GID}" "${USER}"
+if ! id -u "${USER}" ; then
+  echo "creating user [arm] with id ${UID}"
+  useradd --shell /bin/bash \
+    -u "${UID}" -g "${GID}" -G video,cdrom \
+    -o -c "" "${USER}"
+  chown "${USER}.${USER}" "${HOME}"
+  chmod ug+rwX "${HOME}"
+fi
+
+# setup needed/expected dirs if not found
+SUBDIRS="media media/completed media/raw media/movies encode logs db Music"
+for dir in $SUBDIRS ; do
+  thisDir="${HOME}/${dir}"
+  if [[ ! -d "${thisDir}" ]] ; then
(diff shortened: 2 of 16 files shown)


flammableliquids:
It would be epic to simply be able to pull a docker image of this project, it would make set-up a whole lot easier.
You have done a lot of work here, and a lot of it looks really cool. I'm going to learn a lot about docker reading through what you have done.

I wouldn't want to lock users into only using the docker version. Any idea's how we can make it work both ways. Maybe a named fork?

deekue:
I reverted the local changes(oops).  The remaining changes should be a no-op if you're not running in Docker, though they could probably be cleaner.

flammableliquids:
let me know when it's at a point for me to review

scripts/docker-entrypoint.sh
```diff
@@ -0,0 +1,50 @@
...
+
+set -euo pipefail
+
+UID="${UID:-1000}"
+GID="${GID:-1000}"
+export USER=arm
+export HOME="/home/${USER}"
+
+echo "creating group [arm] with id ${GID}"
+groupadd -fo -g "${GID}" "${USER}"
```
supercoffee, line 13: Is there some reason that these operations can't be done in the Dockerfile?  Based on my knowledge of Docker, it seems like it would be more efficient to set up users and groups, and import the config files during build time.

deekue: it depends on how you're consuming this, as a Dockerfile you build into an image yourself or a Docker image you pull from a repo.
this method allows for the user/group to be specified at container start, which makes an image in a repo usable by anyone and for the user/group to be changed without requiring a new image.

deekue:
> let me know when it's at a point for me to review

decided to clean up the changes to the main code.  almost done

kentora:
A few notes, based on how I would use this image (As an image I pull, not one I build, this is my preferred way):

- Would be nice to be able to specify makemkv app_Key as a ENV variable (removes the need to mount a volume for the key only)
- Right now it seems that only a single /dev/sr* is used in the docker instance. Is there any reason not to make 5 or 10 of them?
- Right now the output dirs are placed under the /home/arm/ directory, where the configuration file is also stored. It would be beneficial to bind dir this, but users might want to mount different binds for movies, raw, unidentified and music.

1337-server:
Has anyone been able to get this to rip ?

The ui works but thats all. Are there some instructions im missing ?

arm/ripper/utils.py
```diff
@@ -194,6 +196,53 @@ def make_dir(path):
...
+            }
+    disctype == "unknown"
+    label = udev_dict.get("ID_FS_LABEL", None)
+    if label is None:
+        label == "unknown"
+    elif label == "iso9660":
+        disctype = "data"
+    else:
+        # check for known disctypes
+        for key in map_udev_disctype.keys():
```
1337-server, line 231: I couldn't get Music discs to ever hit this code. I had to rewrite this to get it working with music CD's.

This also isnt helpful for debugging should anything ever go wrong. As it outputs nothing to the user.

arm/ripper/utils.py
```diff
@@ -194,6 +196,53 @@ def make_dir(path):
...
+def parse_udev(udev_dict):
+    map_udev_disctype = {
+            "ID_CDROM_MEDIA_BD": "bluray",
+            "ID_CDROM_MEDIA_DVD": "dvd",
+            "ID_CDROM_MEDIA_TRACK_COUNT_AUDIO": "music"
+            }
+    disctype == "unknown"
+    label = udev_dict.get("ID_FS_LABEL", None)
+    if label is None:
+        label == "unknown"
```
1337-server, line 226: this should be `label = "unknown"` ? The == should break python

scripts/docker_arm_wrapper.sh
```diff
@@ -0,0 +1,112 @@
...
+    disctype="unknown=1"
+  fi
+
+  echo "Starting rip" | logger -t ARM
+  docker exec -i \
+    -u "${ARM_UID}" \
+    -w /home/arm \
+    "${CONTAINER_NAME}" \
+    python3 /opt/arm/arm/ripper/main.py \
+      -d sr0 -t ${disctype} ${label_flag}  \
```
1337-server, line 93: Why doesnt this use  `${DEVNAME}` with small changes to main.py ? This could be made to run easily and would allow the end user to have multiple drives all in one container

1337-server (commented):
Amazing work! It makes things so much easier.

With well written documentation for easy access it will be a fantastic docker/container!

arm/ripper/main.py
```diff
@@ -375,16 +377,21 @@ def main(logfile, job):
     args = entry()
 
     devpath = "/dev/" + args.devpath
```
1337-server, line 379: Removing the "/dev/" + from here would allow you to push the device( ${DEVNAME} ) in from docker_arm_wrapper.sh. Allowing for multiple drives within a single container.

1337-server:
Closing as a full release of docker/container of A.R.M is almost ready to be released.

PR #363, opened 2020-10-26
Update to Master for fixes.

flammableliquids:
Read comments for full details.
Added functionality to curtail number of concurrent handbrake jobs.
Added a logging fix to not pump to the same log file.
Updated readme instructions to be a little more functional, and also to enable nvenc support, plus refer to the correct repo.

PR #366, opened 2020-11-12
Updates to Master for logging improvements and HandBrake limiter

flammableliquids:
Read comments for full details.
Added functionality to curtail number of concurrent handbrake jobs. Modified to separate function.
Added a logging fix to not pump to the same log file.
Updated readme instructions to be a little more functional, and also to enable nvenc support, plus refer to the correct repo.
Modified taking into account the comments of @colonelpopcorn  comments in #363 pull request

### DIFF
--- a/arm/ripper/handbrake.py
+++ b/arm/ripper/handbrake.py
@@ -29,19 +29,7 @@ def handbrake_mainfeature(srcpath, basepath, logfile, job):
     """
     logging.info("Starting DVD Movie Mainfeature processing")
     logging.debug("Handbrake starting: " + str(job))
-    
-    # Wait until there is a spot to transcode
-    job.status = "waiting_transcode"
-    logging.info("Setting job status to 'waiting_transcode' and checking to see if there is a slot")
-    transcode_number = 9999999
-    while transcode_number >= job.config.MAX_CONCURRENT_TRANSCODES:
-        #Count how many times handbrake is currently running
-        #procs = subprocess.check_output(['ps', '-a', '-c', '-ocomm=']).splitlines()
-        #TRANSCODE-NUMBER = procs.count('HandBrakeCLI')
-        transcode_number = sum(1 for proc in psutil.process_iter() if proc.name() == 'HandBrakeCLI')
-        logging.info("Number of HandBrakeCLI Processes running is:" + str(transcode_number) + " going to waiting 12 seconds.")
-        time.sleep(12)
-    
+    utils.SleepCheckProcess("HandBrakeCLI",int(job.config.MAX_CONCURRENT_TRANSCODES))
     logging.debug("Setting job status to 'transcoding'")
     job.status = "transcoding"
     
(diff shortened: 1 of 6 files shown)


arm/ripper/utils.py
```diff
@@ -60,6 +62,18 @@ def scan_emby(job):
...
+def SleepCheckProcess(ProcessStr, Proc_Count):
+
+    if Proc_Count != 0:
+        Loop_Count = Proc_Count + 1
+        logging.debug("Loop_Count " + str(Loop_Count))
+        logging.info("Starting A sleep check of " + str(ProcessStr))
+        while Loop_Count >= Proc_Count:
+            Loop_Count = sum(1 for proc in psutil.process_iter() if proc.name() == ProcessStr)
+            logging.debug("Number of Processes running is:" + str(Loop_Count) + " going to waiting 12 seconds.")
+            time.sleep(10)
```
colonelpopcorn, line 74: Not blocking, but should probably be configurable.

arm/ripper/handbrake.py
```diff
@@ -6,6 +6,9 @@
 import subprocess
 import re
 import shlex
+import time
+import datetime
+import psutil
```
colonelpopcorn, line 11: I'm not familiar with Python enough to know if this import is necessary, but it doesn't seem to be used in the changes you've added. Is this because `utils.SleepCheckProcess` on line 32 depends on `psutil`?

arm/ripper/logger.py
```diff
@@ -23,9 +25,22 @@ def setuplogging(job):
...
     else:
-        logfull = cfg['LOGPATH'] + "/" + logfile
+        #Check to see if file already exists, if so, create a new file
+        TmpLogFull = cfg['LOGPATH'] + "/" + logfile
+        if os.path.isfile(TmpLogFull):
+            #CTime = round(time.time() * 100)
+            logfile = str(job.label) + "_" + str(round(time.time() * 100)) + ".log"
+            logfull = cfg['LOGPATH'] + "/" + logfile
+        else:
+            logfull = cfg['LOGPATH'] + "/" + logfile
```
colonelpopcorn, line 43: Not blocking, but this if else statement looks like it could be consolidated to a ternary statement for the TmpLogFull string.

colonelpopcorn (approved):
Looks good to me! I left a few comments for updates in the future, but it's not blocking!

arm/ripper/utils.py
flammableliquids: Which part? The Duration of the sleep command? The reason I spaced it out to 10 ms was to reduce compute cycles.

arm/ripper/handbrake.py
flammableliquids: Remanent from when I had the function built in. Will remove in next release.

arm/ripper/logger.py
1337-server: This doesnt check if it's the empty.log .This caused my some logging issues and would break the arm empty.log

I had to add a if statement to avoid having a broken log files
`    if logfile != "empty.log":`

flammableliquids: oh snap. I was wondering what happened to that. @133-server, do you want to include this fix, in your upcoming pull request? I have to do one for a blu-ray fix, so I am happy to copy and paste the above code into mine.

1337-server: Sorry i didnt see the notification for this.

Its up to you.  Ive been having alot of issues with the logging tbh. i have changed some of the code, I'm trying to change the logging to have a consistant logging when the script fails. In my experience; ARM if it has issues will just die with no logging. and i have to try running the script via ssh. Im trying to have it so that the NAS log becomes the main log file if there's any problems.

The apprise notifications kind of go around this issue, but im trying to find an easier solution for those who dont have apprise setup.

PR #379, opened 2020-12-20
Security fix for logfiles with api/secret keys

1337-server:
By default the config class throws out everything when a str() of itself is requested.
This small change disables putting out any of the keys that should be kept secret/not posted in log files.


1337-server:
I've put this in its own pull as i think it should be looked into as soon as possible.

PR #412, opened 2021-03-18
some quick fixes

phirestalker:


arm/ripper/main.py
```diff
@@ -92,7 +92,7 @@ def check_fstab():
     with open('/etc/fstab', 'r') as f:
         lines = f.readlines()
         for line in lines:
-            if re.search(job.devpath, line):
+            if re.search(job.devpath, line) and not line.startswith('#'):
```
1337-server, line 95: Wouldn't it be better to use regex since it is already being used for the re.search anyway ? 
I used this is my own fork
`if re.search("^" + job.devpath, line):`

phirestalker: That is much cleaner since it only uses one if statement. Thanks.

Issue #375, opened 2020-12-15
Wiki Update


1337-server:
### Description
I have made an updated version of the wiki [here](https://github.com/1337-server/automatic-ripping-machine/wiki)

### Environment
Ubuntu & Debian/OMV
ARM vers 2.2_dev

I made them as the front readme seemed rather cluttered. Creating the wiki allows the readme to be made much smaller and easier to read.

1337-server:
@flammableliquids This may be worth merging into the ARM wiki.

PR #413, opened 2021-03-21
Adding new Features/Bugfixes

1337-server:
I'm not really sure how to submit all these changes(there is nowhere to discuss large changes), and as I've had no replies the to wiki update question #375 . I think the main repo may be dead. 

So i will leave the changes here as a draft should anyone want to use/add them individually.
The change list is HUGE and breaking these up into smaller pull requests would be a real headache, but if this is needed, its something i can do. Should the main repo become active again.

Any feedback would be great.

# Full list from newest to oldest
   - Improved ajax requests for jobs on front page
   - Abandon job now kills the PID (no files are removed)
   - Added chown to ARMui (Fix permissions button)
   - Bypass for forgotten password (WIP)
   - Bugfix (arm_wrapper.sh was not executable)
   - Bugfix (AttributeError when getting failed/successful jobs)
   - Bugfix (sometimes there is no date in a release from mb)
   - Now added the possibility for users to use [TMDB](https://developers.themoviedb.org/3/getting-started/introduction) as their metadata provider.
  - Added method to let users send/submit their correctly identified movies to a new crc64 API (an api key required)
 - Added check for crc64 from remote database 
 - Updated jquery tablesorter, old version was vulnerable to XSS 
 - Removed all unused versions of CSS 
 - Smalls validation checks when searching the database page. (searching now requires min 3 chars)
 - Small changes to index.html (home for arm ui) to warn before abandoning jobs
 - Jquery ui now fully removed. Now uses only bootstrap for theming
 - Some progress on converting all string types to fstrings where possible
 - ARM will now break out of wait if the user inputs/searches for movie/series.
 - Adding bypass for db.session.commit() error for movies (WIP only MakeMKV and part of handbrake is coded)
 - Abandon job option added to main ARM ui page (for now this only sets job to failed no processes are cancelled)
 - Typo fixes (ARM ui images has/had some typos these have been updated and corrected)
 - ARM ui now shows CPU temps
 - ARM ui now uses percentage bars to more clearly display storage and RAM usage 
 - ARM ui database page is now fully functional with updated ui that looks clearer and with more details
 - ARM ui database is now searchable
 - ARM ui settings page now fully works and saves your settings to the arm.yaml
 - ARM ui now prevents logfiles that contain "../"
 - ARM ui login page updated to look smoother look
 - Bugfix (ARM will no longer log failures when opening sdx or hdx devices)
 - Bugfix (ARM will no longer crash when dealing with non utf-8 chars)
 - Bugfix (ARM database_updater() was setting incorrect values into the database)
 - Bugfix (ARM ui will no longer crash when trying to read logs with non utf-8 chars)
 - Bugfix (ARM ui will now get the latest % of encode, this was buggy as it was getting the very first % it could find)
 - Bugfix (ARM ui will now correctly display ram usage)
 - Bugfix (ARM ui now correctly deals with setup (it should now create all necessary folders and do so without errors) )
 - Bugfix (ARM ui update title no longer shows html on update)
 - Travisci/flake8 code fixes
 - github actions added
 - Bugfix(small bugfix for datadiscs)
 - Bugfix (old versions of yaml would cause Exceptions)
 - Bugfix (db connections are now closed properly for CD's)
 - added bypass for music CD's erroring when trying to commit to the db at the same time
 - A smaller more manageable update this time 
 - Early changes for cleaner looking log output
  - Security (HandBrake.py outputs the new prettytable format)

  - Bugfix (Bluray disc with no titles will now be handled correctly and will not throw an exception )
  - Bugfix (abcde.config now correctly uses musicbrainz)
 - Added prettytables for logging
 - Remove api/keys from the Config class string printout 
    - Added dynamic webserver ip to notifications
    - Allow Deleting entries from the db (also warning for both the page and every delete)
    - Added music CD covers (provided by musicbrainz & coverartarchive.org)
    - Added CPU/RAM info on index page
    - Added some clearer display for the listlogs page 
    - Bugfix (Mainfeature now works when updating in the ui) 
    - Bugfix (Job is no longer added twice when updated in ui) 
    - ALPHA: Added ARM settings page (This only shows settings for the moment, no editing)


flammableliquids:
It's gonna take me ages to review this. But I will. If it kills me.
Then I will merge it to master.

1337-server:
I think part of the reason this looks so big is that it's working against v2.1 instead of the changes from v2.2 that have been already committed to master. There are still a huge number of changes, just but not as big as it initially seems.
I made the pull against branch 2.1 because i know its not used 

Just a couple of small notes 
- All of the issue \# numbers are wrong. They are for their own issues on my fork
- This removes travis.ci in favor of github actions
- Because fstrings are now used everywhere, this locks in users to using a min version of python 3.6 (This was already required, but it can now be made clear to the end user)
- lots of the recent commits have been refactoring to make ARM easier to read and to follow Code Complexity rules. With the end result being ARM going from F > B https://codeclimate.com/github/1337-server/automatic-ripping-machine Still lots of work to do, but it is improving!

PR #427, opened 2021-05-15
Allow password as a parameter.

deisterhold:
Accept a password as a parameter for the Debian silent install script.


1337-server (commented):
Great changes!

1337-server:
Already added

PR #429, opened 2021-05-19
Cleanup README and BASH scripts and automate MakeMKV install.

TechnologyClassroom:
Changes in this PR:

- Automated installation of latest MakeMKV.
  - Added parsing for the latest MakeMKV version instead of relying on hardcoding version numbers.
  - Added checksum verification for MakeMKV download.
- Cleaned up the README.
  - Reordered headings in the README to describe the project and how it works first.  Installation steps follow starting with automated install, then manual install, and then upgrade.
  - Cleaned up references to version 2.2 which does not exist in this repo.  Fixes some of this issue: https://github.com/automatic-ripping-machine/automatic-ripping-machine/issues/393
- Minor edits to README and BASH scripts.
  - Added exit on error to BASH scripts.
  - Changed `apt-get` to `apt`.
  - Added or moved `-y` flags to every `apt` command before or after `install` so further edits can be done with vertical editing.
  - Added periods to sentences.
  - Added a space before sentences in comments.
  - Removed blank space at the end of lines if I came across it.
  - Added carriage returns to make the README markdown easier to read.
  - Added `-p` to `mkdir` commands so they do not fail if the directory already exists.

Issue #423, opened 2021-04-18
Ripping finishes successfully + Want to skip transcode = Everything gets deleted in the end?


nreymundo:
### Description
Hello! Wasn't sure if this was the proper way to request some troubleshooting help but couldn't find any forum or chat link.

I installed ARM into a new VM yesterday and after getting all the setup working I'm really struggilng with a weird behavior now. Here's what happens: 

1. I pop a BluRay disk on.
2. The udev rule correctly picks it up and starts ARM who then correctly identifies it as a movie. 
3. It starts ripping and I see MakeMKV output files appearing in the raw directory. I can even play them and all is great. 
4. At the same time it also creates an empty folder with the movie name in the final movie directory. 
5. Once everything is finished the disc pops out and the logs say everything was moved to the final movies directory correctly.

What actually happens though is that instead of the files getting moved they get deleted from the raw folder (even though I have the retain raw flag to true) and they are not moved to the final place. In fact, the folder with the movie name under `/movies` also gets deleted. 

I'm redirecting all the ARM directories to a couple mounted SMB shares but that shouldn't be an issue as I can see it creates files and I mounted everything so it belongs to the `arm` user. Here's how I'm mounting everything anyway just in case you guys spot an error: 

```
# NAS stuff
//192.168.1.5/media	/mnt/media		cifs	uid=1001,gid=1001,credentials=/home/arm/.smb,iocharset=utf8,noperm 0 0
//192.168.1.5/ingest	/mnt/ingest		cifs	uid=1001,gid=1001,credentials=/home/arm/.smb,iocharset=utf8,noperm 0 0

# Optical drives
/dev/sr0  /mnt/dev/sr0  udf,iso9660  users,noauto,exec,utf8  0  0
```

```
arm in vm-makemkv in ~ 
❯ ls -las /mnt/
total 12
4 drwxr-xr-x  5 root root 4096 Apr 18 00:30 .
4 drwxr-xr-x 20 root root 4096 Apr 17 19:39 ..
4 drwxr-xr-x  3 root root 4096 Apr 18 00:30 dev
0 drwxr-xr-x  2 arm  arm     0 Apr 18 00:48 ingest
0 drwxr-xr-x  2 arm  arm     0 Mar 27 19:34 media
```

I've tried with 2 different blurays and several times with those two with the results being always the same. Any ideas where I might be screwing up?. 

### Environment

```
arm in vm-makemkv in ~ 
[...]

Issue #411, opened 2021-03-15
Ability to cancel a job from the web interface or command line


phirestalker:
### Description
I would like to be able to cancel a job and delete all intermediate files. I tried to go too fast and the 3rd rip stopped with a message letting me know it would take more storage than I had available. It seemed to be waiting indefinitely for me to answer yes or no on whether to continue.

I tried to just kill the PID listed in the web interface for the job, but it is still listed, and I don't know what is going to happen with it.

A button on each job to cancel, or a command I can fire off in the terminal would cancel the job, and delete it from the interface and disk would both work.

### Environment
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.2 LTS"
v2_master
commit df4be42090c6976526d737c7f82802bf5f5ba715 (HEAD -> v2_master, origin/v2_master, origin/HEAD)

1337-server:
If the job is still showing in the web interface, it will be removed the next time you run ARM (manually or from disc insert) if its PID isn't found.

1337-server:
I didn't think to mention it earlier, but after seeing your pull request it might be worth mentioning. I have added this feature in my own repo. There are also a lot of other features added that might be useful or interesting to you.

phirestalker:
I'm glad you mentioned it. I actually visited your repo yesterday. I wasn't sure how to ask you, as I did not want to open a new issue for it.
How easy would it be for me to switch? I know the git commands are trivial, but how compatible is your latest release with this v2_master branch?  Are there any special upgrade instructions?

1337-server:
Not really anything special you need to do to upgrade. 

Just update the files locally and then either insert a disc or run the ARM run command and ARM will do almost all of the rest of the work. 
The only extra work you will need to do is create an admin account so you can delete jobs or edit the arm settings from the Ui.

If you have a look at the wiki on my own repo( I am waiting for a response about adding these to the wiki here) there are lots of extra info about setup, upgrading , changing the themes, hardware transcoding , etc.

But if you get stuck with anything, let me know and I will try to give you any info I may have missed to detail in the documentation.

phirestalker:
Also, while I have you, does your version setup java with MakeMKV for Blu-ray menus?

1337-server:
[...]

Issue #394, opened 2021-01-29
Add more directory support for placing movies, tv shows, and audio for Plex


SysEngDan:
### Description
New Feature: Add more directory support for placing ripped movies, tv shows, and audio into separate folders to support Plex's ability to detect said libraries automatically.

### Use Case
When ARM is configured to store completed rips of movies, tv shows, and audio into separate folder, such as /var/media/movies, /var/media/tvshows, and /var/media/audio, respectively, and say that /var/media is a mounted NFS share which is also mounted to the same mountpoint on a Plex Media Server, then Plex will automatically be able to detect any newly ripped files.

Note: If there is currently a supported method for this, please advise. I only see a movies directory supported in the latest arm.yaml

1337-server:
This wont help for tv shows, but ARM already uses a different folder for Music. 
The config for the music directory is set in the .abcde.conf and not the arm.yaml

SysEngDan:
> This wont help for tv shows, but ARM already uses a different folder for Music.
> The config for the music directory is set in the .abcde.conf and not the arm.yaml

Well that's good to know, I wasn't aware of that yet. Thanks! Perhaps https://github.com/automatic-ripping-machine/automatic-ripping-machine/issues/395 will support the TV show directory better.

1337-server:
Base on the last sentence in [this comment](https://github.com/automatic-ripping-machine/automatic-ripping-machine/pull/373#discussion_r557835448) 
I'm not sure if a tv directory would be supported or added. Although I could be wrong!

1337-server:
Assuming this is now resolved.

PR #436, opened 2021-06-09
V2.1 dev

flammableliquids:
The code base is drifting. 
V2.1 dev has the most updates, and is therefore the best candidate to merge into the master.
After which other pull requests will need to be reviewd


1337-server:
This already contains an icon for waiting so #428 would no longer be needed.

This readme has had huge changes and passes off a lot of the instructions to the wiki. 
So the main repo wiki would need to be updated to include this info.

This would affect pulls #429 , #428 , #427 and includes a modified version of pull #248

This includes fixes for #423 , #411 , #394

flammableliquids:
I wanted to merge v2.1_dev into the master, as it's the one we are using most.
Unfortunately I did not have the time to review it, and I put two other people as reviewers. 
They did not review. So I will re-do the pull-request and set myself as reviewer. 
I might even just push this to master, as we get more bugs from people pulling that.

PR #460, opened 2021-10-31
Additional logic for TV-Series - subfolders per disk, removes need for duplicate rips to be enabled

ITJamie:
removes need for duplicate rips to be enabled.
allows ripping of multiple disks from the same series at the same time without overwriting of files

Will only rip a series disk once as it will detect the duplicate if it exists

example of folder structure on a tv show with disk titles CAPRICA_DISC1 + CAPRICA_DISC2
```
'/home/arm/media/movies/Caprica (2009–2010)/CAPRICA_DISC1/title_01.mkv'
'/home/arm/media/movies/Caprica (2009–2010)/CAPRICA_DISC1/title_02.mkv'
'/home/arm/media/movies/Caprica (2009–2010)/CAPRICA_DISC2/title_01.mkv'
'/home/arm/media/movies/Caprica (2009–2010)/CAPRICA_DISC2/title_02.mkv'
```

### DIFF
--- a/arm/ripper/main.py
+++ b/arm/ripper/main.py
@@ -185,7 +185,7 @@ def main(logfile, job):
                 hboutpath = os.path.join(job.config.MEDIA_DIR, str(job.title))
         else:
             hboutpath = os.path.join(job.config.ARMPATH, str(job.title))
-        # if tv series include the disk label in name. usually contains DISK1, DISK2 etc. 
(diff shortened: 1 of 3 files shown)


1337-server:
Code base has drifted quite a lot, if you could update it inline with the current version and resubmit, that would be great.

Issue #94, opened 2017-10-08
Distributed encoding


MrTechGadget:
I have been thinking a lot about ARM lately as I am happily using it to automate the daunting task of ripping my DVD library.

The most time consuming part is obviously encoding video.
ARM does a great job of queuing up these tasks to run on a single computer.

I have been thinking of how we could easily "kick this up a notch" by parallelizing this encoding if a user has multiple hosts available.

I found this post about a distributed solution for handbrake which has my gears turning (pun intended, as this solution uses gearman to manage a queue of tasks to assign to workers) [https://tim.purewhite.id.au/2010/09/handbrake-encoding-cluster/](https://tim.purewhite.id.au/2010/09/handbrake-encoding-cluster/)

Here are my thoughts which I would love some feedback on.

Main ARM machine, which is where all the ripping occurs and will manage tasks. Can also host encoding if desired.

Worker processes running as docker containers, making it easy to deploy additional workers to handle additional encoding tasks. The docker containers would utilize network storage for the raw ripped data and output to network storage, so they would only require CPU, memory, and networking. By using docker, they could be run on Synology NAS, ESXI or Xen hosts, Macs, Windows, etc.

Thoughts? I'm just thinking about this, don't have really any knowledge of gearman or anything. Just thinking it would have some amazing potential to speed things along.

If anyone else has done any research in this area I would be happy to hear about it.

ahnooie:
Interesting idea, it's not something I would use since I don't have a cluster of idle computers lying around and the hypervisor which ARM is running under is pretty beefy already.  But I don't see why it couldn't be added to the project if other people would find it useful--probably people running weaker computers would find it useful.  What may be interesting is some process to upload the raw files to AWS, spinning up an EC2 cluster for transcoding (have your video ready in minutes instead of hours) then destroy the cluster once the resulting files have been downloaded... but there's always that risk of a large bill from that one blue-ray that has 100 tracks, and bandwidth is probably an issue.

JSylvia007:
Funny you mention the Blu-Ray issue... I thought something was wrong... I'm ripping The Expendables 2 currently and it apparently has at least 60+ tracks...  I didn't know this was a problem, but could see if I was using some hosted service this would be one hell of a bill... It's been ripping now for 2 straight days.

PR #495, opened 2022-05-04
Add backup_dvd and HANDBRAKE_LOCAL cfg options.

garbled1:
# Description
Bring back old ARM style of using MakeMKV to copy DVD's to the HD prior
to transcode (with option backup_dvd).  Add a HANDBRAKE_LOCAL option.  With these two, I can run
ARM semi-distributed via Slurm.  It also allows you to do something like:
HANDBRAKE_CLI: ssh hostname HandBrakeCLI
If you just want to do the transcode on a single remote host.

This opens the door to potentially being able to do fully distributed transcoding, for now, it just does 1 disc per remote host.  This is basically step 1 in order to do #94  I suspect step 2 is to pull in the python slurm library, which may or may not be far too heavyweight.

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Installed slurm_client on host with bluray player.  Built 2 VM's with a slurm cluster.  Set HANDBRAKE_CLI to "srun --exclusive HandBrakeCLI" and RIPMETHOD to "backup_dvd".  Insert 2 discs, and it distributed the transcoding into the cluster.  (with generous amounts of NFS sprinkled in).

- [ ] Ubuntu 18.04
- [ ] Ubuntu 20.04
- [X] Debian 10
- [X] Debian 11
- [ ] Docker

# Checklist:

- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings

### DIFF
--- a/arm/ripper/main.py
+++ b/arm/ripper/main.py
@@ -271,7 +271,7 @@ def main(logfile, job):
         # or
         # dvd with MAINFEATURE off and RIPMETHOD mkv
         hb_in_path = str(job.devpath)
-        if job.disctype == "bluray" or (not cfg["MAINFEATURE"] and cfg["RIPMETHOD"] == "mkv"):
+        if job.disctype == "bluray" or (not cfg["MAINFEATURE"] and cfg["RIPMETHOD"] == "mkv") or cfg["RIPMETHOD"] == "backup_dvd":
             # send to makemkv for ripping
             # run MakeMKV and get path to output
             job.status = "ripping"
--- a/arm/ripper/makemkv.py
+++ b/arm/ripper/makemkv.py
@@ -52,7 +52,7 @@ def makemkv(logfile, job):
     rawpath = setup_rawpath(job, os.path.join(str(cfg["RAW_PATH"]), str(job.title)))
 
     # rip bluray
-    if cfg["RIPMETHOD"] == "backup" and job.disctype == "bluray":
+    if (cfg["RIPMETHOD"] == "backup" or cfg["RIPMETHOD"] == "backup_dvd") and job.disctype == "bluray":
         # backup method
         cmd = f'makemkvcon backup --decrypt {cfg["MKV_ARGS"]} ' \
               f'-r disc:{mdisc.strip()} {shlex.quote(rawpath)}>> {logfile}'
--- a/docs/arm.yaml.sample
+++ b/docs/arm.yaml.sample
@@ -173,11 +173,12 @@ CHOWN_GROUP:
 ## MakeMKV Parameters ##
 ########################
 
-# Method of MakeMKV to use for Blu Ray discs.  Options are "mkv" or "backup".
-# backup decrypts the dvd and then copies it to the hard drive.  This allows HandBrake to apply some of it's
+# Method of MakeMKV to use for Blu Ray discs.  Options are "mkv",  "backup" or "backup_dvd".
+# backup decrypts the bluray and then copies it to the hard drive.  This allows HandBrake to apply some of it's
 # analytical abilities such as the main-feature identification.  This method seems to offer success on bluray 
-# discs that fail in "mkv" mode. *** NOTE: MakeMKV only supports the backup method on BluRay discs.  ARM does
-# not use MakeMKV for DVDs.   
+# discs that fail in "mkv" mode.
+# *** NOTE: MakeMKV only supports the backup method on BluRay discs.
+# backup_dvd forces arm to extract the dvd with MakeMKV prior to the Handbrake step
 RIPMETHOD: "backup"
 
 # MakeMKV Arguments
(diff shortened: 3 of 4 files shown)


1337-server:
> Bring back old ARM style of using MakeMKV to copy DVD's to the HD prior to transcode

ARM already does this. -> [Stacking Discs](https://github.com/automatic-ripping-machine/automatic-ripping-machine/wiki/FAQ#arm-wont-eject-the-dvd-until-it-finishes-transcoding)

But the slurm looks interesting, I'll try this out over the weekend.

garbled1:
Right, but if I set the options the way the faq suggests, then I get the issue with bluray's being ripped as mkv rather than the backup method.  What I want is DVD's ripped as mkv for "stacking" and blurays ripped as backup, which is what the patch implements.

The HANDBRAKE_LOCAL change, is because the initial track counting scan requires access to the disk, so if you srun that, it explodes.  This would also be needed for a simple ssh setup to just do remote encoding.

Quick hint for slurm, make sure the uid/gid of your arm user matches on all nodes, and all paths are the same.  (I just make the media dir a symlink to NFS storage)

Basically, srun --exclusive starts a job, that finds a node it can use exclusively.  So assuming a simple 2 node setup, if you run 5 srun --exclusives, the first 2 launch into nodes, and the next 3 block waiting for a node to free.  As soon as it does, job 3 continues, and so on.  You can also set them to log to separate files, etc etc.  So in theory, you could just load 1 disc in, with say 24 tracks, and have it just srun the whole set, and as nodes free up, it slots handbrakes in.  If you tracked the job # from each, you could then run an srun that set a run condition of after all the other ones finished, and do finalization and block completion on that execution.  In theory this can all be done via the python API, but I haven't investigated yet.

1337-server:
Just an update to say I haven't gotten round to testing this yet… I will get to it as soon as I can.

garbled1:
No hurry on this, just a minor fix found in my torture test of it this weekend.    I threw 4 blurays and 2 dvd's at it as fast as the drive could rip them, and then just let it do it's thing.  It basically interleaved all the work between the 2 remote nodes, and even the webui remained more or less functional.  (the last track to complete just appears stuck at 99% until it gets requeued).  It's still running, but this is nearly perfect for doing a bunch of disks at once, and still having my desktop CPU untouched.

flammableliquids:
None of the changes I see in these files looks prohibitive.
1337-server, if your happy, I am happy to merge the request. (I have been a bit swamped with personal life to stay on top of things)

docs/arm.yaml.sample
```diff
@@ -212,8 +213,10 @@ HB_PRESET_BD: "HQ 1080p30 Surround"
 # Extension of the final video file
 DEST_EXT: mkv
 
-# Handbrake binary to call
+# Handbrake binary to call.  These should be the same unless you are doing remote
+# processing.
 HANDBRAKE_CLI: HandBrakeCLI
+HANDBRAKE_LOCAL: HandBrakeCLI
```
1337-server, line 219: Adding `HANDBRAKE_LOCAL` to the arm.yaml means that it will need to be added to the comments.json file with comments, or some of the armui might break.

Adding this
```
"HANDBRAKE_LOCAL": "# Handbrake binary to call.  These should be the same unless you are doing remote processing.",
```
to the comments file should fix it.

docs/arm.yaml.sample
```diff
@@ -173,11 +173,12 @@ CHOWN_GROUP:
...
 ########################
 
-# Method of MakeMKV to use for Blu Ray discs.  Options are "mkv" or "backup".
-# backup decrypts the dvd and then copies it to the hard drive.  This allows HandBrake to apply some of it's
+# Method of MakeMKV to use for Blu Ray discs.  Options are "mkv",  "backup" or "backup_dvd".
+# backup decrypts the bluray and then copies it to the hard drive.  This allows HandBrake to apply some of it's
 # analytical abilities such as the main-feature identification.  This method seems to offer success on bluray 
-# discs that fail in "mkv" mode. *** NOTE: MakeMKV only supports the backup method on BluRay discs.  ARM does
-# not use MakeMKV for DVDs.   
+# discs that fail in "mkv" mode.
```
1337-server, line 179: Changes to these comments will need to be made to the comments.json

arm/ripper/makemkv.py
```diff
@@ -52,7 +52,7 @@ def makemkv(logfile, job):
     rawpath = setup_rawpath(job, os.path.join(str(cfg["RAW_PATH"]), str(job.title)))
 
     # rip bluray
-    if cfg["RIPMETHOD"] == "backup" and job.disctype == "bluray":
+    if (cfg["RIPMETHOD"] == "backup" or cfg["RIPMETHOD"] == "backup_dvd") and job.disctype == "bluray":
```
1337-server, line 55: This works, but it will cause issues when merging my own fork. But these can be fixed at the time.

arm/ripper/main.py
```diff
@@ -271,7 +271,7 @@ def main(logfile, job):
         # or
         # dvd with MAINFEATURE off and RIPMETHOD mkv
         hb_in_path = str(job.devpath)
-        if job.disctype == "bluray" or (not cfg["MAINFEATURE"] and cfg["RIPMETHOD"] == "mkv"):
+        if job.disctype == "bluray" or (not cfg["MAINFEATURE"] and cfg["RIPMETHOD"] == "mkv") or cfg["RIPMETHOD"] == "backup_dvd":
```
1337-server, line 274: This works, but it will cause issues when merging my own fork. But these can be fixed at the time.

1337-server (changes requested):

garbled1:
Sorry for the delay, got busy with a different issue.  Will get this cleaned up as requested asap.

1337-server:
Could you please resubmit this against the v2_devel branch ?

Thanks

Issue #658, opened 2023-01-14
Ref #657, Ubuntu install script still broken


aiohdfgaiuhg:
- [X] I have checked the wiki
- [X] I have searched for similar issues

**Describe the bug**
Breaking hot-fix 2.6.10 unfortunately did not fix the issue.  I am now seeing a more simple error. 

when running script, I see 
failed due to runui.py.  When I run that python script manually i get the following. 

/opt/arm/arm/runui.py: 1: Main run page for armui: not found
import-im6.q16: attempt to perform an operation not allowed by the security policy `PS' @ error/constitute.c/IsCoderAuthorized/408.
import-im6.q16: attempt to perform an operation not allowed by the security policy `PS' @ error/constitute.c/IsCoderAuthorized/408.
/opt/arm/arm/runui.py: 6: Syntax error: word unexpected (expecting ")")

### Environment
Ubuntu 20.04.5LTS in Proxmox VM, new install
2.6.10 hot fix

shitwolfymakes:
Closing as duplicate of #592 

I'll try and duplicate this later, but I've never had this issue come up. May need to reimage and reinstall

PR #655, opened 2023-01-13
2.6.9 Hotfixes

shitwolfymakes:
# Description
Fixes for breakages in the ubuntu installer script. This should be merged ASAP to prevent users from using the broken installer

This PR patches the following issues:
- #654  
- Fixes launch crash loop caused by the new copy destination of abcde.conf  

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [X] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.
Please also list any relevant details for your test configuration

- [ ] Ubuntu 18.04
- [x] Ubuntu 20.04
- [ ] Debian 10
- [ ] Debian 11
- [ ] Docker
- [ ] Other (Please state here)

# Checklist:

- [x] **I have submitted this PR against the `v2_devel` branch (Unless the main branch has a security issue)**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have tested that my fix is effective or that my feature works

Issue #592, opened 2022-10-17
Web UI does not start; Cannot verify if ARM is fully installed.


J-Farmer:
- [X] I have checked the wiki
- [X] I have searched for similar issues

**Describe the bug**
The UI never launches once ARM is installed. Running the runui.py file manually errors out on both Ubuntu Desktop and Server.

**To Reproduce**
1. Install based on automatic setup in wiki. ARM user is set up at install time.

2. Ui never launches. systemctl status armui.service shows the following: Process: 53440 ExecStart=/usr/bin/python3 /opt/arm/arm/runui.py (code=exited, status=1/FAILURE)

3. Running the runui.py file at /opt/arm/arm errors out with: 
 ```
./runui.py: 1: Main run page for armui: not found
import-im6.q16: attempt to perform an operation not allowed by the security policy \`PS' @ error/constitute.c/IsCoderAuthorized/408.
import-im6.q16: attempt to perform an operation not allowed by the security policy \`PS' @ error/constitute.c/IsCoderAuthorized/408.
./runui.py: 6: Syntax error: word unexpected (expecting ")")
```

### Environment
[OS Distribution and version (run "cat /etc/lsb-release")]
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.5 LTS"

[ARM Release Version or if cloning from git branch/commit (run "git branch" and "git log -1" to get this)] 
This should be running the latest release based on the instructions in the wiki.

I had this working in a VM on Proxmox, and was attempting to set it up on bare metal so I could have access to more DVD drives at once. I did run into some similar issues starting this on Proxmox, but despite errors, the web ui would start and I could rip movies. I can't on a dedicated machine.

J-Farmer:
One correction from my post: I set up the arm user when installing the OS, not at arm install time.

shitwolfymakes:
I would recommend trying to reinstall ARM, which simply requires running the script again. I've never seen this issue before. I have no idea why an ImageMagick error is occurring, as this project does not have it in the dependencies.

Also, it would be helpful if you could fix the error log, please put all relevant text into a \`\`\`...\`\`\` code block (you may need to escape some tilda's to make it work)

Two questions:
1. Was Firefox already running in the VM when you ran the installer script
2. Did you do anything other than create a default installation with the `arm` user?

[...]

Issue #587, opened 2022-10-06
can't get to setup page on auto installer on 20.04


bjquinn:
- [X] I have checked the wiki
- [X] I have searched for similar issues

**Describe the bug**
First-time setup never launches when running automatic setup.

**To Reproduce**
1. Follow instructions for automatic setup for Ubuntu 20.04
2. First time setup never launches
3. Manually loading http://[ip]:8080/setup in the browser results in "Unable to connect Firefox can’t establish a connection to the server"

### Environment
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.5 LTS"

arm@fireflash:/opt/arm$ git branch
* (no branch)
arm@fireflash:/opt/arm$ git log -1
commit 66b784e95437ab38e81425ce06f86f7e9afcc578 (HEAD, tag: v2.6.7, origin/v2_devel)
Author: wolfy <wolfy@shitwolfymakes.com>
Date:   Mon Sep 26 23:58:18 2022 -0600

    dev branch now properly checked out during clone (#580)

There doesn't appear to be any errors in the console when installing.  A few warnings like the following:

WARNING: The script apprise is installed in '/home/arm/.local/bin' which is not on PATH.

bjquinn:
Oh ok awesome, thx.  So all I need to do is insert a disc and it should get started?

shitwolfymakes:
You might want to fiddle with the settings to get them to your liking, but otherwise yeah.

Per the wiki, the default username and password are both `arm`

bjquinn:
Ok thanks.  I went ahead and put in a disc and out a few times to see what would happen and I couldn't see anything in empty.log.

But I saw that sudo systemctl status armui was saying the following:

```
Oct 06 21:02:55 fireflash python3[107671]: FileNotFoundError: [Errno 2] No such file or directory: '/etc/arm/config/abcde.conf'
Oct 06 21:02:55 fireflash systemd[1]: armui.service: Main process exited, code=exited, status=1/FAILURE
Oct 06 21:02:55 fireflash systemd[1]: armui.service: Failed with result 'exit-code'.
```

So I did the following, and that seemed to fix it:

`ln -s /etc/.abcde.conf /etc/arm/config/abcde.conf`

[...]

PR #660, opened 2023-01-14
Installer Hotfixes

shitwolfymakes:
More hotfixes for the installer

# Description

A typo from when I copied the patches from my test environment created a bug in #655. This occurred because the correct files already existed in the test environment, so when testing the erroring code the bug did not manifest.

This also adds:
- A small bugfix to prevent an issue where grep not finding a running address for ARM would trigger pipefail and crash the script (this is due to grep-specific returns, not a failure in the logic)
- A small QoL change. The script now waits 30 seconds to give the service ample time to start itself before grepping for the address. This should reduce issues where setup launch fails even though ARM starts successfully, because it launches before systemd completes starting the service.

Fixes # (issue title here)
- #658 
- #592 
- #587 

## Type of change

# How Has This Been Tested?

- [ ] Ubuntu 18.04
- [x] Ubuntu 20.04
- [ ] Debian 10
- [ ] Debian 11
- [ ] Docker

# Checklist:

- [x] **I have submitted this PR against the `v2_devel` branch (Unless the main branch has a security issue)**


scripts/installers/ubuntu-20.04-install.sh
```diff
@@ -248,7 +248,9 @@ function install_armui_service() {
 
 function launch_setup() {
     echo -e "${RED}Launching ArmUI first-time setup${NC}"
-    site_addr=$(sudo netstat -tlpn | awk '{ print $4 }' | grep ".*:${PORT}")
+    echo "Giving ArmUI a moment to start, standby..."
+    sleep 30
```
1337-server, line 252: Didn't I already add this ?

1337-server (approved):

scripts/installers/ubuntu-20.04-install.sh
shitwolfymakes: I thought I added it as part of 2.6.9 but I guess I forgot to copy it over. I'm basing off of v2_devel as of like an hour ago too

1337-server (approved):

PR #738, opened 2023-02-16
Workflow hotfix

shitwolfymakes:
This should fix the broken test runs


1337-server:
What does this fix ? The build seems to be fine at the moment.

1337-server:
I think this was because Dependabot has different secrets to the 'main' repo

Not sure if this is still needed ?

shitwolfymakes:
This PR fixes the workflow `PR - Test Docker Build`, which currently errors on every PR (except this one)

PR #751, opened 2023-02-22
Updates to MakeMKV key handling and extras

### LINKED ISSUE #626: MakeMKV failed with code: 2 (DVD ejects a few minutes after insertion)
- [x] I have checked the wiki
- [x] I have searched for similar issues

**Describe the bug**
I installed ARM via the `docker-setup.sh` script and followed the instructions to run the container with the provided script. It starts and I am able to interact with the web interface. When I insert a DVD, the job starts but after a few minutes, when the ripping process begins, the DVD makes some noises and then ejects the disk with the error below. I tried different DVDs but it's the same for each of them.

Any ideas?

**To Reproduce**
1. Follow docker installation instructions
2. Instert DVD

### Environment
[OS Distribution and version (run "cat /etc/lsb-release")] Debian 11
[ARM Release Version or if cloning from git branch/commit (run "git branch" and "git log -1" to get this)] 
"latest" Docker images: `automaticrippingmachine/automatic-ripping-machine   latest    cde3a25b2c07   5 weeks ago   1.59GB`

ARM Git version: 47b8e540b555e1cfc1287a5a49a4f032476bfa8d

### Log file
[Run the rip in DEBUG and drag and drop the log file onto this comment box]

[>>>>> DRAG AND DROP LOG FILE HERE <<<<<]
[ANGELS_SHARE.log](https://github.com/automatic-ripping-machine/automatic-ripping-machine/files/10158584/ANGELS_SHARE.log)

### LINKED ISSUE #752: 🔧 arm.yaml migration

- [X] I'm using the latest release

Currently arm.yaml is not overwritten when updates are installed, to prevent overwriting user settings. This presents a problem: when arm.yaml is changed as part of the version, those changes do not trickle down to current users. We need a way to update arm.yaml while preserving the current user settings.

At present, we have a similar case already handled. Whenever the user saves changes to the settings, the new data is written to the file.

The logical loop for this new functionality should be as follows:
During STARTUP:
1. Load `/etc/arm/config/arm.yaml` (the user's current settings) and `/opt/arm/setup/arm.yaml` (The version from github)
2. If the dictionaries DO NOT have the same number of keys, then
3. Update the values of keys in the new dict with the old values
4. Copy `/opt/arm/setup/arm.yaml` to `/etc/arm/config/arm.yaml`, destructively
5. Save the dictionary containing current user settings + new options to `/etc/arm/config/arm.yaml`

This is really the only way to do it while preserving user settings

### Anything else?

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

shitwolfymakes:
# Description

This PR adds the ability to set a permanent MakeMKV key to disable key updating.

## Type of change

# How Has This Been Tested?

Tested all inputs and outputs of `update_key.sh`

- [x] Docker

# Checklist:
- [x] **I have submitted this PR against the `main` branch (Unless the main branch has a security issue)**

# Changelog
- Added `MAKEMV_PERMA_KEY` to `arm.yaml`, which stores the user's permanent key
- Resolves #752
- Fixes `PR - Test Docker Build` action breaking on PRs from forks
- `arm_user_files_setup.sh` no longer erroneous creates a directory called "0777"
- `ui/utils.arm_yaml_check_groups()` and `ui/utils.arm_yaml_test_bool()` have been moved into the new `config/config_utils.py`
- `update_key.sh` now takes in the permanent key as an argument, and only scrapes the beta key if it runs
- `update_key.sh` now checks to see if the `.MakeMKV` dir exists, and creates it if not
- `update_key.sh` now either appends or replaces the beta or permanent key in `settings.conf`, as appropriate. This allows MakeMKV settings to persist.
- Updated the PR template to reflect recent changes
- Reinstated `docker-setup.sh`, which was accidentally deleted
- Removed `MAKEMKV_APP_KEY` env variable from the Docker container
- `Bump version` workflow now only runs on pushes and PRs to `main`
- `Quality Checks` workflow now only runs on pushes and PRs to `main`
- Dropped 3.6.0 quality check (EOL Python release)
- Added `cancel-in-progress` attribute to the `Qualit Checks`, `Publish Docker Image`, `Run Shellcheck`, and `PR - Test Docker Build` workflows to help limit notification spam
- Removed `Test Debian installer` workflow
- Reorganized Dockerfile, significantly reducing image build times
- General housekeeping across all modified files
- Fixed #626 

# Logs:
[POINT_BREAK_167713144185.log](https://github.com/automatic-ripping-machine/automatic-ripping-machine/files/10811053/POINT_BREAK_167713144185.log)


shitwolfymakes:
Test rip is ongoing. Feel free to review, but please don't merge until I've confirmed the rip succeded

shitwolfymakes:
Log attached. Ready for review and merge

arm/config/config.py
```diff
@@ -15,15 +18,55 @@ def _load_config(fp):
...
+# 1. Load both current and template arm.yaml
+cur_cfg = _load_config(arm_config_path)
+new_cfg = _load_config("/opt/arm/setup/arm.yaml")
+
+# 2. If the dicts do not have the same number of keys
+if len(cur_cfg) != len(new_cfg):
+    # 3. Update new dict with current values
+    for key in cur_cfg:
+        if key in new_cfg:
+            new_cfg[key] = cur_cfg[key]
```
1337-server, line 38: Somewhere around here it converts all of the `true` to 1 and 'false' to 0 in the arm.yaml and then saves it

This isnt really a huge issue if they are passed as ints but if they are treated as strings then they are all true. Also it may be confusing for some users.
Will need fixed to show `true` or `false` again before this can be merged.

scripts/docker-setup.sh
```diff
@@ -0,0 +1,94 @@
...
+    echo -e "${RED}Pulling image from $IMAGE${NC}"
+    sudo -u arm docker pull "$IMAGE"
+}
+
+function setup_mountpoints() {
+    echo -e "${RED}Creating mount points${NC}"
+    for dev in /dev/sr?; do
+        sudo mkdir -p "/mnt$dev"
+    done
+    sudo chown arm:arm /mnt/dev/sr*
```
1337-server, line 74: Is this script to be run with or without sudo ?
If its run as sudo then there should be no need to add the sudo here ? Or should sudo be added to line 60 ?

scripts/docker-setup.sh
```diff
@@ -0,0 +1,94 @@
...
+}
+
+function save_start_command() {
+    url="https://raw.githubusercontent.com/automatic-ripping-machine/automatic-ripping-machine/v2_devel/scripts/docker/start_arm_container.sh"
```
1337-server, line 78: the `v2_devel` branch depreciated and is no longer writeable. It should link to main instead.

shitwolfymakes: Ah, fixed that in the test environment but never committed

1337-server: Ok, just ping when its ready for re-review.

scripts/docker-setup.sh
shitwolfymakes: Ah, this is just a copy and paste of the old script

shitwolfymakes:
Discovered the issue here. In the current version, the following code reads in arm.yaml into a dict, preserving the boolean typing.
https://github.com/automatic-ripping-machine/automatic-ripping-machine/blob/1047dfbdc049378ebdbe043c7b743cf23d40e345/arm/config/config.py#L14

However, when the settings are saved, the settings are saved to a dictionary, with the values being string types
https://github.com/automatic-ripping-machine/automatic-ripping-machine/blob/1047dfbdc049378ebdbe043c7b743cf23d40e345/arm/ui/settings/settings.py#L110

Which is why the tests for type are located in `build_arm_cfg()`:
https://github.com/automatic-ripping-machine/automatic-ripping-machine/blob/1047dfbdc049378ebdbe043c7b743cf23d40e345/arm/ui/utils.py#L655-L660

So now that the same logic is being used to do the arm.yaml migration, the old checks produce different results. 
So, `post_value = int(value)` resulted in a `ValueError` because `value` is of type `String`, when coming from the text boxes in the settings page. 

But when my new config.py code runs, `post_value = int(value)` results in a `post_value` storing `1` because `value` is `True` and is of type `bool`, when coming directly from `safe_load()`.

Tl;dr: copying the old code is why it broke, but once I figured out why, I though it was worth documenting for future us.

arm/config/config.py
shitwolfymakes: See my comment below

shitwolfymakes:
Ready for review once tests complete

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,22 @@
+name: "Run ShellCheck"
```
1337-server, line 1: Would it not be better going with the more popular `ludeeus/action-shellcheck` ?

1337-server (changes requested):

.github/workflows/shellcheck.yml
shitwolfymakes: This is the one that popped up first when I was looking for shellcheck actions, and has a higher cadence of development

PR #775, opened 2023-03-02
Fix for #758 - Container needs --privileged flag to work

### LINKED ISSUE #758: :bangbang: Container currently requires --privileged flag to run correctly :bangbang:
This is caused by
https://github.com/automatic-ripping-machine/automatic-ripping-machine/blob/1047dfbdc049378ebdbe043c7b743cf23d40e345/arm/ripper/identify.py#L30

Containers don't have enough privileges to mount, and without this arm fails/makes a mistake here as the mountpoint will be empty

https://github.com/automatic-ripping-machine/automatic-ripping-machine/blob/1047dfbdc049378ebdbe043c7b743cf23d40e345/arm/models/models.py#L133-L145

Because it cant access the filesystem it will automatically revert to data disc type, even if it had previously gotten the correct disc type from here 

https://github.com/automatic-ripping-machine/automatic-ripping-machine/blob/1047dfbdc049378ebdbe043c7b743cf23d40e345/arm/models/models.py#L103-L106

I have already created a working fix for this issue, but will need more time to test it locally. Then i will submit a PR.

1337-server:
# Description
This removed the second check if running inside a container.
Containers cannot use the mount command leading to a failure to detect the correct disc type.
The container should have already found the correct type anyway.

Fixes #758 

## Type of change

# How Has This Been Tested?

- [x] Docker

# Checklist:

# Changelog:

- updated mount command to warn if not mounted
- added if to skip trying to search for mounted files if not mounted

# Logs 
RIPMETHOD:
- backup: 
  - [WRATH_OF_THE_TITANS_167772071829.log](https://github.com/automatic-ripping-machine/automatic-ripping-machine/files/10866982/WRATH_OF_THE_TITANS_167772071829.log) (with SYS_RAWIO enabled)
  - [WRATH_OF_THE_TITANS_167772055090.log](https://github.com/automatic-ripping-machine/automatic-ripping-machine/files/10866986/WRATH_OF_THE_TITANS_167772055090.log) (with SYS_RAWIO disabled)
- mkv: 
  - [WRATH_OF_THE_TITANS_168037865379.log](https://github.com/automatic-ripping-machine/automatic-ripping-machine/files/11130536/WRATH_OF_THE_TITANS_168037865379.log) (with SYS_RAWIO enabled)
  - [WRATH_OF_THE_TITANS_168038083367.log](https://github.com/automatic-ripping-machine/automatic-ripping-machine/files/11130580/WRATH_OF_THE_TITANS_168038083367.log)(with SYS_RAWIO disabled)
- Audio CD: [Leona-Lewis-Spirit.log](https://github.com/automatic-ripping-machine/automatic-ripping-machine/files/10867480/Leona-Lewis-Spirit.log)


1337-server:
It looks like removing the --privileged flag seriously hurts performance when reading from the drive.
In my testing it doubles the time required for ripping, performance when encoding isnt effected. Only accessing the drive, however adding 
`--cap-add SYS_RAWIO` to the docker run command seems to fix this issue.

arm/ripper/identify.py
```diff
@@ -20,17 +19,32 @@
...
         os.makedirs(str(job.mountpoint))
-
-    os.system("mount " + job.devpath)
-
-    # Check with the job class to get the correct disc type
-    job.get_disc_type(utils.find_file("HVDVD_TS", job.mountpoint))
+    # Check and mount drive - log error if failed
+    mounted = check_if_mounted(os.system("mount " + job.devpath))
+    # get_disc_type() checks local files, no need to run unless we can mount
+    if mounted:
```
shitwolfymakes, line 45: Pretty sure this is a fatal error. You'll want the function to throw an error, and call it from a try/catch that'll kill the job if the error is caught

1337-server: Making this a fatal error would mean that running arm in container would never be possible without the `--priv` flag.

shitwolfymakes: The `--priv` that the user needs to get rips to mount the disks in the first place, yeah

1337-server: But its not needed.... The rip logs are attached. In the initial comment
Ripped and converted successfully without the `--privileged` flag.

Because ARM only mounts the disc to search for the disctype and to calculate the CRC64 which are only used by the dvd side of things.
The rest seem to rip without any issue, and the DVD only fails sometimes with the auto ident because its using the crc64. But if it can match omdb all is good.

Will need a lot more testing but removing the `--privileged` flag would be a big plus for the security side of things.

shitwolfymakes: Gotcha. I'm all for eliminating that risk if it's possible. Does `SYS_RAWIO` negate the performance loss too, or does it just fix the issue?

1337-server: I've only tested it with a USB drive, didn't have time for a full test. But it removed the performance issues.

It still ripped no problem without it, it just went from 13/15 mins fully ripped up to 23/25. Wouldn't make a difference using ARM normally (inserting disc and walking away) but for dev time it makes a huge diff in performance.

shitwolfymakes: Gotcha. I would say when you're ripping your entire library, 10 minutes per disk is game changing

shitwolfymakes:
Please include log(s) for running with `mkv`

1337-server:
> Please include log(s) for running with `mkv`

Done!

shitwolfymakes (approved):

PR #863, opened 2023-06-19
Update docker-setup.sh to detect if 'start_arm_container.sh' exists and rename it during docker setup

### LINKED ISSUE #862: 🔧 do not over-write "start_arm_container.sh" if it already exists

- [ ] I'm using the latest release

When updating to pull down the latest docker image, I didn't realize it would be wiping out my "start_arm_container.sh" script.  It's a minor annoyance to fill it back out, but it would be nice if it detected a file existed there and didn't overwrite it.  Alternatively, if the format of this file is likely to change, I'd love if the install script would rename the old one.

existing "start_arm_container.sh" script is somehow preserved

_No response_

### Anything else?

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

TallPatrick:
Added rudimentary handling for checking if "start_arm_container.sh" already exists, and changing the name to "start_arm_container_OLD.sh" if it does, so the new "start_arm_container.sh" can be written

# Description
Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. 
List any dependencies that are required for this change.

Fixes #862 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?
Tested locally by running the new code snippet.

- [x] Docker

# Checklist:

- [n/a] I have made corresponding changes to the documentation

# Changelog:

- Single check in, detailed above.

# Logs
N/A

### DIFF
--- a/scripts/docker-setup.sh
+++ b/scripts/docker-setup.sh
@@ -83,7 +83,7 @@ function save_start_command() {
     if [ -e start_arm_container.sh ]
     then
         echo -e "'start_arm_container.sh' already exists.  Renaming old file to 'start_arm_container_OLD.sh'"
-        sudo mv /start_arm_container.sh ./start_arm_container_OLD.sh
+        sudo mv /start_arm_container.sh ./start_arm_container.sh.bak
     fi
     sudo -u arm curl -fsSL "$url" -o start_arm_container.sh
     chmod +x start_arm_container.sh
--- a/scripts/docker-setup.sh
+++ b/scripts/docker-setup.sh
@@ -80,6 +80,11 @@ function setup_mountpoints() {
 function save_start_command() {
     url="https://raw.githubusercontent.com/automatic-ripping-machine/automatic-ripping-machine/main/scripts/docker/start_arm_container.sh"
     cd ~arm
+    if [ -e start_arm_container.sh ]
+    then
+        echo -e "'start_arm_container.sh' already exists.  Renaming old file to 'start_arm_container_OLD.sh'"
+        sudo mv /start_arm_container.sh ./start_arm_container_OLD.sh
+    fi
     sudo -u arm curl -fsSL "$url" -o start_arm_container.sh
     chmod +x start_arm_container.sh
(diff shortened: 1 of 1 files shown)


microtechno9000:
if there is a script in the folder, why download, you could skip the download altogether

scripts/docker-setup.sh
```diff
@@ -80,6 +80,11 @@ function setup_mountpoints() {
 function save_start_command() {
     url="https://raw.githubusercontent.com/automatic-ripping-machine/automatic-ripping-machine/main/scripts/docker/start_arm_container.sh"
     cd ~arm
+    if [ -e start_arm_container.sh ]
+    then
+        echo -e "'start_arm_container.sh' already exists.  Renaming old file to 'start_arm_container_OLD.sh'"
```
shitwolfymakes, line 85:
```suggestion
        echo -e "'start_arm_container.sh' already exists. Backing up..."
```

scripts/docker-setup.sh
```diff
@@ -80,6 +80,11 @@ function setup_mountpoints() {
...
+        sudo mv /start_arm_container.sh ./start_arm_container_OLD.sh
```
shitwolfymakes, line 86:
```suggestion
        sudo mv /start_arm_container.sh ./start_arm_container.sh.bak
```

shitwolfymakes (changes requested):
The requested changes are the commonly accepted way of doing this operation

scripts/docker-setup.sh
TallPatrick: yeah, this is a better solution.  I'm too used to personal projects where additional jank like this is normal for me.

TallPatrick:
@microtechno9000 
> if there is a script in the folder, why download, you could skip the download altogether

Backing up the existing file is generally the safer option in case there is ever a breaking change with "start_arm_container.sh" that is needed all of the values set by users will be retained in the backed-up old version of the script.

shitwolfymakes (approved):
Great work! Thanks for your help

PR #877, opened 2023-07-04
Docker image generalisation

doudalidou:
Hello,

I tried for so many time to create that kind scripts/system, Thank you very much !!
I really like your Webui that a excellent idea !

The docker part of the image that was online was a bit too hard to understand and debug. 
I've work on it for 3 days and manage to not change your code too much.

My goal was to simplify and make your Docker works for everyone:
- Dockerfile rebuild to include systemd-core and requirements
- Some file name inconsistencies (.abcde vs abcde, some log files) corrected
- Add a "EJECT_DISK" config to let user remove the dvd by them self (and for testing purpose)
- Volume & files structures have been rework to fit with the Docker container
- Add a new library for DVD crc64 because the first one ignores lowercase "TS" directories 

My next steps would be:
- Reorganize configurations so user would be able to change them from their container at first start.
- Upgrade the home page to be able to start a RIP from the web page 
- Upgrade the Jobs detail so we will be able to follow with track is currently ripping
- Upgrade the Jobs detail so we will be able to skip the manual information lookup

Ultimately (and I don't know if I will have the time), I would like to use HttpSocket instead of refreshing the page.

I wanted to let you all want I did, I will continue on my side to upgrade everything I see.

Again, thank you so much for your work that so helpfull for me !!


doudalidou:
I found a problem with the VOLUME, I'LL be back with working one.

Issue #642, opened 2023-01-04
ARM should not use the flask development server


fbermel:
The webui is currently served by the flask development webserver. This is not advisable and the content should be served by a production ready wgsi server like for example waitress.

To switch from flask dev server to waitress the following could be done:

Append the this to `requirements.txt`
```
waitress==2.1.2
```
Install requirements with pip

Replace these lines (26-27) in `runui.py`:
```
26	    app.run(host=host, port=cfg['WEBSERVER_PORT'], debug=True)
27	    # app.run(debug=True)
```

with this
```
26	    from waitress import serve
27	    serve(app, host=host, port=cfg['WEBSERVER_PORT'])
```

Waitress should be a drop in replacement and thus far through my testing everything seems to be working as expected.

shitwolfymakes:
We currently accept the risk for running the development webserver, do to our requirement to be able to restart the webserver during runtime. 

That being said, I've never liked that due to the glaring security risks. Feel free to try get this running with a more secure option, and then PR it with a rundown for us!

shitwolfymakes:
Closing as stale

PR #964, opened 2023-10-24
Use Waitress as WSGI Server

anthonylavado:
# Description
Switches the main web/WSGI server from the Flask development server to Waitress.

Fixes # (issue title here)
#642

## Type of change

# How Has This Been Tested?

- [x] Docker

Tested in Docker, with a fork of arm-dependencies that installs Waitress 2.1.2 (current stable).

# Checklist:

# Changelog:

- Switch from Flask development server to Waitress

# Logs
Currently ripping disc as proof of concept. Will attach logs later.

Corresponding PR in arm-dependencies: https://github.com/automatic-ripping-machine/arm-dependencies/pull/234


microtechno9000:
what does waitress do, differently to Flask Dev?

anthonylavado:
Flask itself includes a basic WSGI server in the Werkzeug component, sufficient for development needs. It is, to quote their documentation, "not designed to be particularly efficient, stable, or secure."

Instead, they recommend using a "production WSGI server", a provide a list of many options. These servers are focused on one task, and they do it well. Of the list, Waitress was chosen as their example because it is easy to configure, which is probably why the issue reporter chose it. It also doesn't require any more dependencies, so it is an easy drop in replacement.

Even if ARM installs are not being exposed to the web it makes sense to use better components. It could be more stable for existing users,  and more secure in the long run.

References:
[Run with a Production Server](https://flask.palletsprojects.com/en/3.0.x/tutorial/deploy/#run-with-a-production-server)
[Deploying to Production](https://flask.palletsprojects.com/en/3.0.x/deploying/)

anthonylavado:
An update from my testing:

I've done various things with it so far (rip a disc, perform title searches, click on every button/link, change settings, change theme). Everything seems to be working as expected, with no issues.

There's nothing particularly interesting in the logs, but I can still provide them if you want.

shitwolfymakes:
I know one of the reasons we've stuck with the Flask dev server for so long, despite the gaping security vulnerabilities, is because that was how @1337-server was able to add a restart feature. Does that still function properly/is it still relevant?

microtechno9000:
@anthonylavado pulled down your fork and built the docker image from the waitress branch. I get the below error when rebuilding the docker image (using devtools). ARM doesn't run and the docker log shows it trying to start and failing, as per below

```
[2023-10-27 10:32:42,149] INFO ARM: runui.<module> Starting ARMUI on interface address - 172.17.0.2:8080
Traceback (most recent call last):
  File "/opt/arm/arm/runui.py", line 44, in <module>
    from waitress import serve
ModuleNotFoundError: No module named 'waitress'
Starting web ui
```

anthonylavado:
> ModuleNotFoundError: No module named 'waitress'

@microtechno9000 You'd need to make sure waitress is available to the container, which would require the update to arm-dependencies that I've linked (and then to build a new image of that, etc).

If you wanted to add it manually here (for testing, or just to do it separately), add this to the dockerfile:
`RUN pip install waitress==2.1.2`


I would put it at line 69, after the COPY for arm.

@shitwolfymakes I think the restart button was the only one I didn't try. Everything else worked though! Will try after the run for the label PR.

microtechno9000:
> I know one of the reasons we've stuck with the Flask dev server for so long, despite the gaping security vulnerabilities, is because that was how @1337-server was able to add a restart feature. Does that still function properly/is it still relevant?

Can't see any issues with Waitress, recommend approving

@shitwolfymakes  and @1337-server , the restart button appears to do nothing in either flask or waitress. Inspecting traffic from the browser, I can't see any queries sent back to the server, nor any logs in docker showing the restart was called.

Issue #949, opened 2023-09-20
🔧 Make URL in notifications configurable without using SERVER_IP and SERVER_PORT


koeberlue:
### Is this feature missing in the latest version?

- [X] I'm using the latest release

### Is your feature request related to a problem? Please describe.

I am running arm inside a docker container where I am mapping the web ui port to a different host port.

```
docker run -p "8000:8080" ...
```

I have also configured Apprise to send me notifications (via Telegram, in case this matters).

When I insert a movie into my disc drive, I get the following notification:

> **ARM notification**
> Found disc:
> Some-Movie. Disc type is dvd [...] Edit entry here: http://127.0.0.1:8080/jobdetail?job_id=<id>

So obviously this link does not work as it points to localhost instead of my arm machine. Also the port is different to the one that is exposed within my local network.

### Describe the solution you'd like?

I would like to set a variable, e.g. `UI_BASE_URL` (I can't think of better name right now).

Then I could set `UI_BASE_URL="http://arm.home:8080"`. 

Whenever I receive a notification, it would use this variable (if set):

> **ARM notification**
> Found disc:
> Some-Movie. Disc type is dvd [...] Edit entry here: http://arm.home:8000/jobdetail?job_id=<id>

### Describe alternatives you've considered?

* Correctly configure `SEVER_IP` and `SERVER_PORT` to construct this URL
  * I could not get this running because the `SERVER_IP` is used as listen address as well

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #972, opened 2023-11-02
Add configurable URL for notifications

anthonylavado:
# Description
In #949, a user proposes adding a configurable field that would be used for notifications. This adds a field called `UI_BASE_URL` for that purpose.

Fixes # 949

## Type of change

# How Has This Been Tested?

- [x] Docker

I get notifications through Telegram, and I have all my containers behind a reverse proxy (Caddy) for ease of access at home. With this in place and configured, the notification URL brings me directly to the ARM page to review the job.

# Checklist:

# Changelog:

- Add UI_BASE_URL to arm.yaml, for setting a display address in notifications

# Logs
Full logs can be provided, but here's an example of the result:
image

# Notes

I can add documentation if required.

I have not added any type of input validation on this field, but it would clearly be suited to entering a protocol and hostname. An enhancement would be to add two extra items:
1. If no protocol is entered, automatically append "http://" to the beginning of the field's contents
2. (Maybe) perform filtering to ensure only a hostname is entered, not spurious text
These could likely be both satisfied by implementing [urllib.parse](https://docs.python.org/3.0/library/urllib.parse.html).


anthonylavado:
The SonarCloud fail is kind of funny actually, since it already had http in the original notification, I just moved it around.

Forgot to mention - I'm open to changing the name if anyone has a better suggestion. The only one I could think of was `display_address` which is what I use in the code itself. This was also my first time working with Alembic, so I'm also open to adding any documentation you want around that (to [here](https://github.com/automatic-ripping-machine/automatic-ripping-machine/tree/main/arm/migrations)).

microtechno9000:
@anthonylavado can you add a comment/close the sonarlint review. Had a look and it is just around http vs https. ARM isn't using https so no issues, can mark the security issue as closed/safe. Doesn't look like I can do that.

anthonylavado:
@microtechno9000 It doesn't look like I can change it either. It might only be set up to allow Wolfy to do so.

I'll fix the merge conflict before we tag him.

1337-server:
The issue should now be ok. I marked it as safe.

microtechno9000:
> @microtechno9000 It doesn't look like I can change it either. It might only be set up to allow Wolfy to do so.
> 
> I'll fix the merge conflict before we tag him.

once the merge conflict is resolved, I can review and close. No issues with the modification

anthonylavado:
I think I've pulled in the changes to models and fixed it, but it looks like a right mess. I'm just building locally to test and make sure it still works.

anthonylavado:
Looks like we're good. Thanks all!

microtechno9000 (approved):

PR #1125, opened 2024-05-10
Fix permissions issue caused by a8a8b8f

shitwolfymakes:
# Description
Fixed an issue created by removing permissions setting in commit a8a8b8f. Confirmed working after running the rebuilt container.

The wiki now includes a note about default permissions for newly created folders and how that make affects users on mounting file shares

## Type of change

# How Has This Been Tested?

- [x] Docker

# Checklist:


shitwolfymakes:
@SylvainMT This will fix your issue once merged and released

microtechno9000 (commented):
Agree with the fix, seems the removal of chown to resolve a specific issue was not worth the general hassle.

Can you amend the wiki to explain that the default behaviour sets the user ARM as the owner using chown. 

There are users that have mounted the ARM folders via a samba or NFS share that this will cause issues. Need to make sure those users are aware and mounting network shares is not the preferred config.

shitwolfymakes:
I see what you mean now. The way this is fixing the permissions issue will break permissions for shares every time it boots. I'll figure out a different way to fix this.

shitwolfymakes:
@SylvainMT This will no longer automatically fix your issue. If you are not using an external share for the container, running this should fix the permissions issues:
```
sudo chown -R arm:arm /home/arm
```

microtechno9000 (approved):
good with the proposed changes, best of both worlds

SylvainMT:
I also agree, that permissions need only be set at folder creation, not every time the container starts.  Thanks for looking into the issue :)

PR #1290, opened 2025-01-24
Update intel-qsv.md

simonmcnair:
clarification of the processors that support QSV

I have a i5-11400F and spent ages trying to get this to work assuming it was like SSE or MMX and just an extended cpu instruction set.  I didn't know it required a iGPU

I got informed, and thought it would be wise to spread information in the hope of reducing issues.


SylvainMT:
I think it is a good addition to add a link to Intel`s Processor list to verify if there is an integrated GPU. That being said, Intel nomenclature says that if the F is present in the processor's name then no iGPU is present.  

Most motherboard vendors will also allow disabling the iGPU in the bios.  That is another reason why some may have trouble with this.  I think that would be worth mentioning.

microtechno9000 (approved):

PR #1304, opened 2025-02-08
[FEATURE] Automate parts of the default start docker command script

wapolinar:
# Description
Computes uid and gid of user arm dynamically in the start script
Use a default number of cores (all cores -1)  in the start script

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

# How Has This Been Tested?
Tested start script by following the new, edited documentation

- [X] Docker

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings

# Changelog:

- Auto-configuration for arm uid and gid in docker
- Automatic CPU core calculation for docker
- Small bugfix in configuration (core 3 -> threads 5,6)

### DIFF
--- a/scripts/docker/start_arm_container.sh
+++ b/scripts/docker/start_arm_container.sh
@@ -1,4 +1,15 @@
 #!/bin/bash
+set -e
+function join_by {
+  local d=${1-} f=${2-}
+  if shift 2; then
+    printf %s "$f" "${@/#/$d}"
+  fi
+}
+cores_minus_one=`nproc --ignore=1`
+cpu_array=`seq 1 1 $cores_minus_one`
+cpu_cores=$(join_by , ${cpu_array})
+
 docker run -d \
     -p "8080:8080" \
     -e ARM_UID="`id -u arm`" \
@@ -15,5 +26,5 @@ docker run -d \
     --privileged \
     --restart "always" \
     --name "arm-rippers" \
-    --cpuset-cpus='2,3,4,5,6,7...' \
+    --cpuset-cpus="$cpu_cores" \
     IMAGE_NAME
--- a/scripts/docker/start_arm_container.sh
+++ b/scripts/docker/start_arm_container.sh
@@ -25,6 +25,6 @@ docker run -d \
     --device="/dev/sr3:/dev/sr3" \
     --privileged \
     --restart "always" \
-    --name "arm-rippers" \
+    --name "automatic-ripping-machine" \
     --cpuset-cpus="$cpu_cores" \
     IMAGE_NAME
--- a/scripts/docker/start_arm_container.sh
+++ b/scripts/docker/start_arm_container.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 docker run -d \
     -p "8080:8080" \
-    -e ARM_UID="<id -u arm>" \
-    -e ARM_GID="<id -g arm>" \
+    -e ARM_UID="`id -u arm`" \
+    -e ARM_GID="`id -g arm`" \
     -v "<path_to_arm_user_home_folder>:/home/arm" \
     -v "<path_to_music_folder>:/home/arm/music" \
     -v "<path_to_logs_folder>:/home/arm/logs" \
--- a/arm_wiki/docker.md
+++ b/arm_wiki/docker.md
@@ -36,35 +36,25 @@ The script will now:
 5. Save a copy of the template run command for the user to fill in to `~arm/start_arm_container.sh`
 
 ## Post Install
-1. **ARM User ID**: In a terminal session, type `id -u arm` and make a note of the returned value
-2. **ARM Group ID**: In a terminal session, type `id -g arm` and make a note of the returned value
-3. **Start Script**: Open up `start_arm_container.sh` in the text editor of your choice. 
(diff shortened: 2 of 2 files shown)


microtechno9000:
nice work, will avoid a few permission issues hopefully

scripts/docker/start_arm_container.sh
```diff
@@ -14,6 +25,6 @@ docker run -d \
     --device="/dev/sr3:/dev/sr3" \
     --privileged \
     --restart "always" \
-    --name "arm-rippers" \
-    --cpuset-cpus='2,3,4,5,6,7...' \
+    --name "automatic-ripping-machine" \
+    --cpuset-cpus="$cpu_cores" \
```
mihawk90, line 30: You can shorten this whole segment about building your variable and remove the function by using the aforementioned `$()` syntax instead:

```suggestion
    --cpuset-cpus="$(seq --separator=, 1 1 $(nproc --ignore=1))" \
```

This is much cleaner, simpler, and certainly more readable.

As I gather that function is from this SO post:
https://stackoverflow.com/a/17841619/3323286
But it lacks in explaining what it actually does and isn't really readable (and unnecessary to begin with).

---

That aside, if we're going to pass a default number of ~~cores~~ threads, we should also give the option to users to change this default, which can be easily achieved by creating a new variable and passing that to the `nproc` call.

scripts/docker/start_arm_container.sh
```diff
@@ -1,8 +1,19 @@
...
+cores_minus_one=`nproc --ignore=1`
+cpu_array=`seq 1 1 $cores_minus_one`
+cpu_cores=$(join_by , ${cpu_array})
+
 docker run -d \
     -p "8080:8080" \
-    -e ARM_UID="<id -u arm>" \
-    -e ARM_GID="<id -g arm>" \
+    -e ARM_UID="`id -u arm`" \
+    -e ARM_GID="`id -g arm`" \
```
mihawk90: This old-style notation is generally discouraged, use the newer `$()` syntax instead:

```suggestion
    -e ARM_UID="$(id -u arm)" \
    -e ARM_GID="$(id -g arm)" \
```

microtechno9000 (approved):

microtechno9000:
@wapolinar can you please resolve the conflicts with the docker wiki and I can merge

wapolinar:
Sorry for the delay, but updated the branch and also added automatic timezone computation.

arm_wiki/Docker.md
microtechno9000, line 1: Once you have updated the script as an 'automated' script, reference the two auto and manual scripts in the wiki

microtechno9000, line 1: Once you have updated the script as an 'automated' script, reference the two auto and manual scripts in the wiki

scripts/docker/start_arm_container.sh
microtechno9000, line 1: Move this code to a new script 'automated_arm_container.sh` to have a manual and auto container. That will allow users to chose, noting not all users will want an automated script

wapolinar:
I am not sure if that really makes a lot of sense.
The documentation tells the user to use `id -u arm` and `id -g arm` and that is automated (and should work on any Linux distro) - so why this could be useful as a manual step, I am not sure. At least the documentation should be updated to tell the user in which cases this should be something else than the output of the `id` command?

Regarding the `--cpuset-cpu` command, the updated documentation states: 
> By default, all but one core/thread will be used.

And all of the original documentation is retained (so it is possible to change for everyone that is interested).

In the case of `timedatectl show -p Timezone --value` the documentation has been removed completely. I agree that this might be not a good idea, so it could be performed similar to the cpuset command, retain and update the documentation, stating that by default the timezone of the host machine will be used.

Since it is still possible to change every value if needed and the documentation (after update) then also clearly states that, I am not sure if two separate scripts really do make sense?

microtechno9000 (approved):

microtechno9000:
> I am not sure if that really makes a lot of sense.
> The documentation tells the user to use `id -u arm` and `id -g arm` and that is automated (and should work on any Linux distro) - so why this could be useful as a manual step, I am not sure. At least the documentation should be updated to tell the user in which cases this should be something else than the output of the `id` command?
> 
> Regarding the `--cpuset-cpu` command, the updated documentation states: 
> > By default, all but one core/thread will be used.
> 
> And all of the original documentation is retained (so it is possible to change for everyone that is interested).
> 
> In the case of `timedatectl show -p Timezone --value` the documentation has been removed completely. I agree that this might be not a good idea, so it could be performed similar to the cpuset command, retain and update the documentation, stating that by default the timezone of the host machine will be used.
> 
> Since it is still possible to change every value if needed and the documentation (after update) then also clearly states that, I am not sure if two separate scripts really do make sense?

Fair

I can approve, please update the version. The version script isn't working

mihawk90 (commented):
Due to how stupid the GitHub UI for Reviews is, I didn't realise until a couple days ago that a review needs to be "submitted" separately from writing the comment to be visible to everyone ...

OH WELL, it's only been 4 months since I wrote this and apparently never submitted :skull: 

Anyway, here are my comments.

microtechno9000:
closing as stale

Issue #1315, opened 2025-02-28
🐛 Fatal Error: Invalid literal for int() with base 10: 'A1'


drr159:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### I have checked the wiki

- [x] I have searched the wiki for this problem

### Does this issue exist in the latest version?

- [x] I'm using the latest release

### Describe the bug?

I am trying to rip an album (Linkin Park - Meteora 20th Anniversary) and 2 of the 3 discs fail with the same error. The primary audio disc succeeds, but the "LPU Rarities 2.0" and "Live Rarities 2003-2004" discs fail. I originally had this issue on ARM 2.10.3, and upgraded to 2.10.5, but that did not resolve the issue. Attached are the logs from attempting to rip the "LPU Rarities 2.0" disc. Debug logging is enabled. 

Note: when using "freecddb" for the GET_AUDIO_TITLE parameter in arm.yaml, the two failing cds were able to rip successfully. This is only an issue with using the "musicbrainz" option.

### To Reproduce

1. ARM 2.10.5 using musicbrainz for GET_AUDIO_TITLE parameter in arm.yaml
2. Rip one of the discs used above (or possibly others, though I do not currently have other examples)

### What OS are you seeing the problem on?

all

### ARM Release Version or Git hash (You can find this in the arm settings page)

Python Version: 3.10.12
ARM Version: 2.10.5
Current git version: 3f4cd7c
Docker image id: 5ba8e857719b

### Expected behavior?

It should rip fine without any issues

### Relevant log output

```shell
[02-28-2025 14:57:50] DEBUG ARM: music_brainz.music_brainz discid = fmAlyTnA0j.qTjeAVVnJN_2XNuQ-
[02-28-2025 14:57:50] ERROR ARM: main.<module> invalid literal for int() with base 10: 'A1'
Traceback (most recent call last):
  File "/opt/arm/arm/ripper/main.py", line 223, in <module>
    main(log_file, job, args.protection)
  File "/opt/arm/arm/ripper/main.py", line 116, in main
    music_brainz.main(job)
  File "/opt/arm/arm/ripper/music_brainz.py", line 25, in main
    return music_brainz(discid, disc)
  File "/opt/arm/arm/ripper/music_brainz.py", line 52, in music_brainz
    process_tracks(job, infos['disc']['release-list'][0]['medium-list'][0]['track-list'])
[...]

PR #1320, opened 2025-03-03
[BUGFIX] Fix for "invalid literal for int() with base 10: 'A1'"

microtechno9000:
# Description
Bugfix for issue raised #1315 
When ripping a cd, the track number comes back as 'A1' for incorrectly identified discs.
Bugfix runs a string check against the returned number, if a string will then use the position field

```json
...
'track-list':[
    {
        'id':'8ed112e1-af3e-4d5c-b3c8-bcbd4ccab165',
        'position':'1',
        'number':'A1',
        'length':'13440',
        }
...
```

Fixes # (issue title here)

## Type of change

# How Has This Been Tested?

- [ ] Docker
- [x] Other (Pending user feedback)

# Checklist:

# Changelog:

Bugfix to music_brainz.py

# Logs
Pending user feedback

### DIFF
--- a/arm/ripper/music_brainz.py
+++ b/arm/ripper/music_brainz.py
@@ -242,7 +242,15 @@ def process_tracks(job, mb_track_list, is_stub=False):
                 track_leng = int(track['recording']['length'])
         except ValueError:
             logging.error("Failed to find track length")
-        trackno = track.get('number', idx + 1)
+
+        # Get Track ID
+        # Check if track is returned as a string like 'A1' for Vinyl
+        if isinstance(track.get('number', idx + 1), str):
+            trackno = track.get('position', idx + 1)
+            logging.info(f"Using Track 'position' as 'number' <str> track: {trackno}")
+        else:
+            trackno = track.get('number', idx + 1)
+
         if is_stub:
             title = track.get('title', f"Untitled track {trackno}")
         else:


mihawk90:
Mh I wonder if something more is going on here...

> When ripping a cd, the track number comes back as 'A1' for incorrectly identified discs.

I don't think this is wrongly identified at all.

When I search for the DiscID `fmAlyTnA0j.qTjeAVVnJN_2XNuQ-` that was being logged on Musicbrainz, it only returns 2 CDs:
https://musicbrainz.org/cdtoc/fmAlyTnA0j.qTjeAVVnJN_2XNuQ-

1. https://musicbrainz.org/release/4cb5db86-2d27-4106-a6d1-1d8a85b11b20
2. https://musicbrainz.org/release/3483663a-35a4-4a63-a210-bf8362a9c030

Notice on the first release those are 5 LPs + 4 CDs.

According to the annotation that first release was a Limited Edition. The CDs are likely the same pressing as the non-Limited (second) release, which explains the identical DiscID.

So the problem with this fix would be that it results in wrong metadata for the ripped tracks.

On the DiscID-tab of the release we can see that this DiscID is actually CD7 (or CD2, depending how you count):
https://musicbrainz.org/release/4cb5db86-2d27-4106-a6d1-1d8a85b11b20/discids

With the tracklist:
1.	A.06
2.	Pretty Birdy (Somewhere I Belong 2002 demo)
3.	Sold My Soul to Yo Mama
4.	Standing in the Middle
5.	Program (Meteora demo)
6.	Faint (demo 2002)
8. …

Correct me if I'm wrong, but if I'm reading this code correctly, you're just overriding the (lettered) track-numbers (from the LP) with the position numbers. By doing that you're basically accepting the first LP as the source of the track list, which is different:
1. A1	Foreword
2. A2	Don’t Stay
3. A3	Somewhere I Belong
4. A4	Lying From You
5. A5	Hit the Floor
6. A6	Easier to Run
7. …

I think the only way to fix that would be to check against the DiscID for every track :/ From what I can tell from the JSON blob (although prettifiers refuse to parse it so.. :headstone: ), every track has the details on the disc attached as well.

microtechno9000:
@mihawk90 Yes the fix is not the best, but was intended as a 'quick fix'. Nothing like them in production right :p

I didn't see the differences you have pointed out in the JSON data, will have a look and see if there is a better way to get the right information.

PR #1418, opened 2025-06-15
[BUGFIX] Handle missing log directories more gracefully

alwold:
# Description
I recently set up ARM on a bare metal Ubuntu 25.04 install (I'm hoping to do a writeup soon for that as well). Since there is more manual setup involved, I forgot to create the `logs/progress` directory, which caused the `main.py` script to crash inside the log cleanup function. When that happened, the exception wasn't logged in any of the logs, as far as I could tell, and the job was left in the database (and couldn't be abandoned since the PID was no longer is running).

This makes two small changes to better handle this scenario:

- Move the calls to the cleanup function inside the `try` block in `main.py`. That way if an exception is thrown during log cleanup, it is caught, logged, and the job is cleaned up properly.
- If any of the log directories are missing, log something and continue on to the next directory without throwing an exception.

This is my first contribution, and I'm happy to adjust the PR if anything isn't following the correct procedures, just let me know!

## Type of change

# How Has This Been Tested?

- [ ] Docker
- [x] Other (Please state here)
 
I tested these on bare metal since I don't have docker, but it should be possible to test on docker as well, if you can build an image where you remove the `logs/progress` directory.

## Change to log cleanup
- Make sure your `logs/progress` directory does not exist
- Start a rip
- Verify that a message is logged during log cleanup indicating the progress directory was skipped
- Verify that the rip continues

## Change to top level exception handling
- Comment out the change in `logger.py` to continue when a logging directory is missing
- Start a rip
- Verify that the job throws an exception, but it is handled and logged, and the job is properly aborted

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have tested that my fix is effective or that my feature works

# Changelog:

- Handle missing log directories during cleanup in `logger.py`
- Expand the scope of the top level try block in `main.py` so it handles exceptions in log cleanup

# Logs

### DIFF
--- a/arm/ripper/logger.py
+++ b/arm/ripper/logger.py
@@ -84,6 +84,9 @@ def clean_up_logs(logpath, loglife):
     for log_dir in logs_folders:
         logging.info(f"Checking path {log_dir} for old log files...")
         # Loop through each file in current folder and remove files older than set in arm.yaml
+        if not os.path.isdir(log_dir):
+            logging.info(f"{log_dir} is not a directory or doesn't exist. Skipping.")
+            continue
         for filename in os.listdir(log_dir):
             fullname = os.path.join(log_dir, filename)
             if fullname.endswith(".log") and os.stat(fullname).st_mtime < now - loglife * 86400:
--- a/arm/ripper/main.py
+++ b/arm/ripper/main.py
@@ -211,15 +211,15 @@ if __name__ == "__main__":
     with open(os.path.join(cfg.arm_config["INSTALLPATH"], 'VERSION')) as version_file:
         version = version_file.read().strip()
 
-    # Delete old log files
-    logger.clean_up_logs(cfg.arm_config["LOGPATH"], cfg.arm_config["LOGLIFE"])
-    logging.info(f"Job: {job.label}")  # This will sometimes be none
-    # Check for zombie jobs and update status to 'failed'
-    utils.clean_old_jobs()
-    # Log all params/attribs from the drive
-    log_udev_params(devpath)
-
     try:
+        # Delete old log files
+        logger.clean_up_logs(cfg.arm_config["LOGPATH"], cfg.arm_config["LOGLIFE"])
+        logging.info(f"Job: {job.label}")  # This will sometimes be none
+        # Check for zombie jobs and update status to 'failed'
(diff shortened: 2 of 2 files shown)


arm/ripper/main.py
microtechno9000, line 1: The try here is wrapped with the main ARM ripper code, suggest that if this fails the whole job doesn't need to fail.
Add an additional try/exception before the main, that will just throw an error that the log files are missing and move on

arm/ripper/logger.py
```diff
@@ -84,6 +84,9 @@ def clean_up_logs(logpath, loglife):
     for log_dir in logs_folders:
         logging.info(f"Checking path {log_dir} for old log files...")
         # Loop through each file in current folder and remove files older than set in arm.yaml
+        if not os.path.isdir(log_dir):
```
microtechno9000, line 87: no issues with this

microtechno9000:
Thanks for making ARM better, and adding logs to your PR :)

The version script isn't working for some reason at the moment, please bump your ARM version manually

arm/ripper/main.py
alwold: Hi @microtechno9000. Sorry for the delay on this, but I think it should be updated now. It will now catch/log the error and continue on with the job.

I've also bumped the version.

I also think there may still be some value in expanding the scope of the `try:` block that calls `main()`, so that ARM can recover in the case that some earlier part of the job fails in an unexpected way, but I'm not sure where the best place to start it would be...

VERSION
```diff
@@ -1 +1 @@
-2.17.1
\ No newline at end of file
+2.17.2
```
1337-server: Needs latest version

alwold: Ok, I merged in the latest from main and updated the version. Let me know if there's anything further I need to do RE the submodule. It's been a while since I used them.

1337-server (changes requested):
Sorry about the delay, can you please bump the version and we can merge ?

alwold:
> Sorry about the delay, can you please bump the version and we can merge ?

@1337-server ok, it's updated! Sorry for the delay on my end as well :) I'll try to keep it updated if the VERSION in main changes again.

microtechno9000 (approved):

PR #1476, opened 2025-07-23
[FEATURE]UI safety updates Ripper settings yaml and UI

flammableliquids:
# Description

The forms "Ripper Settings" And "UI Settings" Have now been upgraded.
This includes the beginnings of field validatiors and drop downs.
"Ripper Settings" form has too many fields to be defined manually in python, and so a json has been created to allow for ease of organization and management of the form. 

Fixes # (issue title here)

# How Has This Been Tested?
Many many repeated tests in 'bare metal mode'

- [ ] Docker
- [x] Other (24.04 Ubuntu)

# Checklist:

- [ ] I have tested that my fix is effective or that my feature works

# Changelog:

- arm/ui/settings.SettingsForm() is not hard-coded for rendering, if a new field is added to the form, the loop will now render it.
- applied to "UI Settings":
- - DataRequired validator now 
- - Hinting now applied to fields (not just the blue 'i' icon but on hover.
- - use_icons and save_remote_images changed to dropdowns to prevent user mistakes.
- arm/ui/forms.SettngsForm() completely re-created:
- - is no longer a class, but a function that returns an instance of a class.
- - creates an instance an empty class, and then based on ripperFormConfig.json populates the fields(types)
- - Hinting now applied to fields.
- - Dropdown fields and integer fields now enforce user input to prevent users from crashing their servers (This all started because I typed DEBRUG instead of DEBUG and had to plug in to the server to find the mistake). 
- ripperFormConfig.json created, and quickJsonGenerator.py (to create ripperFormConfig.json), so that the field types, and validators for the "Ripper Settings" form can be easily managed via loops instead of massive python code.
- arm/ui/forms_custom_validators.py added. To hold custom validators. Primary goal validator is the "validate_path_exists" to prevent users from inputting files or folders that do not exist on the server. (If you do this, you will cause your server to crash) (validator either does not work, or the rendering of form error fields doesn't show the error)
- improved datatyping in forms.py
- arm/ui/utils.py Updated: Improved dataype indications and function return type indications. Functions modified or added:
- - listCoPairedIntoTuple (Ability in name).
- - jsonFile_to_dict : provided a str|Path will read the file and return a dict.
- - generate_comments modified to use jsonFile_to_dict
- - generate_ripperFormSettings reads ripperSEttingsConfigFile.json and uses jsonFile_to_dict to return a dict
- Jinja Changes:
- - _form_field.html created, containing the marco 'render_field(field)', takes a field, and renders it, including the 'i' popover.
- - arm/ui/settings/templates/settings/ui.html imports render_field,  and instead of manually executing against each field in the form, loops over the passed form, and calls render_feild
- - arm/ui/settings/templates/settings/ripper.html imports render_field,  and instead of manually executing against each field in the form, loops over the passed form, and calls render_feild
- [BugFix] arm/runui.py updated for better path support.
- [BugFix] arm/ui/comments.json was Missing the UMASK Field
- Documentation updates. Including link correction


flammableliquids:
Re-opening after minor description fixes and removing unwanted file.

PR #1487, opened 2025-07-29
[BUGFIX] replace hard-coded paths with configurable ones

xieve:
Adhering to systemd standards for the environment variables.

## Description

All hard-coded paths can now be configured via environment variables. The names of the environment variables are chosen so that they can be read directly from systemd, if arm is started as a systemd service, see [systemd.exec(5)](https://man.archlinux.org/man/core/systemd/systemd.exec.5.en#SANDBOXING).

### Type of change

- [x] This change requires a documentation update

## How Has This Been Tested?
Ripped and transcoded a DVD successfully

- [x] Docker
- [x] Other (NixOS Module)

## Checklist:

## Changelog:

- Configs are read from `$CONFIGURATION_DIRECTORY`
- `$RUNTIME_DIRECTORY` is used for mountpoints
- `$STATE_DIRECTORY` is used for the DB
- `$LOGS_DIRECTORY` is used for the logs
- The default values are as before
- Read `$PATH` from `/etc/environment` in the docker wrapper so we can call makemkv without an absolute path
  - In my opinion, this is the better approach vs. a1df08d1b93ff25926d5983c1f186407ff236c63, because it is more general, that is why I haven't rebased onto main yet

## Logs
[EN-102281_175380153613.log](https://github.com/user-attachments/files/21493557/EN-102281_175380153613.log)

### DIFF
--- a/arm/config/config.py
+++ b/arm/config/config.py
@@ -6,7 +6,7 @@ import yaml
 
 import arm.config.config_utils as config_utils
 
-CONFIG_LOCATION = "/etc/arm/config"
+CONFIG_LOCATION = os.environ.get("CONFIGURATION_DIRECTORY", "/etc/arm/config")
 arm_config_path = os.path.join(CONFIG_LOCATION, "arm.yaml")
 abcde_config_path = os.path.join(CONFIG_LOCATION, "abcde.conf")
 apprise_config_path = os.path.join(CONFIG_LOCATION, "apprise.yaml")
@@ -28,7 +28,13 @@ def _load_abcde(fp):
 # handle arm.yaml migration here
 # 1. Load both current and template arm.yaml
 cur_cfg = _load_config(arm_config_path)
-new_cfg = _load_config("/opt/arm/setup/arm.yaml")
+new_cfg = _load_config(os.path.join(cur_cfg['INSTALLPATH'], "setup/arm.yaml"))
+
+# Environment variables as in systemd.exec(5)
+if log_path := os.environ.get("LOGS_DIRECTORY"):
+    cur_cfg["LOGPATH"] = log_path
+if state_path := os.environ.get("STATE_DIRECTORY"):
+    cur_cfg["DBFILE"] = os.path.join(state_path, "arm.db")
 
 # 2. If the dicts do not have the same number of keys
 if len(cur_cfg) != len(new_cfg):
@@ -38,7 +44,10 @@ if len(cur_cfg) != len(new_cfg):
             new_cfg[key] = cur_cfg[key]
 
     # 4. Save the dictionary
-    with open("/opt/arm/arm/ui/comments.json", "r") as comments_file:
+    with open(
+        os.path.join(cur_cfg["INSTALLPATH"], "arm/ui/comments.json"),
+        "r",
+    ) as comments_file:
         comments = json.load(comments_file)
 
     arm_cfg = comments['ARM_CFG_GROUPS']['BEGIN'] + "\n\n"
--- a/arm/ripper/main.py
+++ b/arm/ripper/main.py
@@ -13,11 +13,15 @@ import time  # noqa: E402
 import datetime  # noqa: E402
 import re  # noqa: E402
 import getpass  # noqa E402
+from importlib.util import find_spec
+from pathlib import Path
 import pyudev  # noqa: E402
 import psutil  # noqa E402
 
-# set the PATH to /opt/arm so we can handle imports properly
-sys.path.append("/opt/arm")
+# If the arm module can't be found, add the folder this file is in to PYTHONPATH
+# This is a bad workaround for non-existent packaging
+if find_spec("arm") is None:
+    sys.path.append(str(Path(__file__).parents[2]))
 
 from arm.ripper import logger, utils, identify, arm_ripper, music_brainz  # noqa: E402
 import arm.config.config as cfg  # noqa E402
--- a/arm/models/job.py
+++ b/arm/models/job.py
@@ -136,7 +136,7 @@ class Job(db.Model):
     def __init__(self, devpath):
         """Return a disc object"""
         self.devpath = devpath
-        self.mountpoint = "/mnt" + devpath
+        self.mountpoint = os.environ.get("RUNTIME_DIRECTORY", "/mnt") + devpath
         self.hasnicetitle = False
         self.video_type = "unknown"
         self.ejected = False
--- a/arm/ripper/ARMInfo.py
+++ b/arm/ripper/ARMInfo.py
@@ -47,7 +47,7 @@ class ARMInfo:
         Function to get the current arm git version
         """
         branch_len = 10
-        cmd = "cd /opt/arm && git branch && git log -1"
+        cmd = f"cd {self.install_path} && git branch && git log -1"
         git_output = ProcessHandler.arm_subprocess(cmd, True)
         git_regex = r"\*\s(\S+)\n(?:\s*\S*\n){1,10}(?:commit )([a-z\d]{5,7})"
         git_match = re.search(git_regex, str(git_output.decode("utf-8")))
--- a/arm/ripper/arm_ripper.py
+++ b/arm/ripper/arm_ripper.py
@@ -4,8 +4,13 @@ It would help clear up main and make things easier to find
 import sys
 import os
 import logging
+from importlib.util import find_spec
+from pathlib import Path
 
-sys.path.append("/opt/arm")
+# If the arm module can't be found, add the folder this file is in to PYTHONPATH
+# This is a bad workaround for non-existent packaging
+if find_spec("arm") is None:
+    sys.path.append(str(Path(__file__).parents[2]))
 
 from arm.ripper import utils, makemkv, handbrake  # noqa E402
 from arm.ui import app, db, constants  # noqa E402
--- a/arm/ripper/logger.py
+++ b/arm/ripper/logger.py
@@ -117,7 +117,9 @@ def create_logger(app_name, log_level=logging.DEBUG, stdout=True, syslog=False, 
 
     if file:
         # create file logger handler
-        file_handler = logging.FileHandler('/home/arm/logs/arm.log')
+        file_handler = logging.FileHandler(
+            os.path.join(cfg.arm_config['LOGPATH'], 'arm.log')
+        )
         file_handler.setLevel(log_level)
         file_handler.setFormatter(formatter)
         logger.addHandler(file_handler)
--- a/arm/ripper/makemkv.py
+++ b/arm/ripper/makemkv.py
@@ -816,10 +816,7 @@ def prep_mkv():
     """
     try:
         logging.info("Updating MakeMKV key...")
-        cmd = [
-            "/bin/bash",
-            "/opt/arm/scripts/update_key.sh",
-        ]
+        cmd = [os.path.join(cfg.arm_config["INSTALLPATH"], "scripts/update_key.sh")]
         # if MAKEMKV_PERMA_KEY is populated
         if cfg.arm_config['MAKEMKV_PERMA_KEY'] is not None and cfg.arm_config['MAKEMKV_PERMA_KEY'] != "":
             logging.debug("MAKEMKV_PERMA_KEY populated, using that...")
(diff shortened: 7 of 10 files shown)


xieve:
Example for a systemd service: (rough sketch, will work later with more changes from my fork)
<details>

```ini
# /etc/systemd/system/armui.service
[Unit]
After=network-online.target
Description=Automatic Ripping Machine Web UI
Wants=network-online.target

[Service]
BindPaths=/mnt/frail/srv/rips/raw/
BindPaths=/mnt/frail/srv/rips/transcoded/
BindPaths=/mnt/frail/srv/rips/completed/
ConfigurationDirectory=arm # /etc/arm
DeviceAllow=block-sr rw
ExecStart=/usr/bin/armui
LogsDirectory=arm # /var/log/arm
LogsDirectory=arm/progress
PrivateTmp=true
ProtectHome=true
ProtectSystem=strict
Restart=always
RestartSec=3
StateDirectory=arm # /var/lib/arm
Type=exec
User=arm

[Install]
WantedBy=multi-user.target
```

```ini
# /etc/systemd/system/arm@.service
[Unit]
Description=Automatic Ripping Machine Worker

[Service]
ConfigurationDirectory=arm # /etc/arm
ExecStart=/usr/bin/arm --no-syslog --devpath "%I"
LogsDirectory=arm
PrivateTmp=true
ProtectHome=true
ProtectSystem=strict
ReadWritePaths=/mnt/frail/srv/rips/raw/
ReadWritePaths=/mnt/frail/srv/rips/transcoded/
ReadWritePaths=/mnt/frail/srv/rips/completed/
ReadWritePaths=/dev/%I
RuntimeDirectory=arm # /run/arm
StateDirectory=arm # /var/lib/arm
User=arm
```

</details>

arm/config/config.py
```diff
@@ -6,7 +6,7 @@
 
 import arm.config.config_utils as config_utils
 
-CONFIG_LOCATION = "/etc/arm/config"
+CONFIG_LOCATION = os.environ.get("CONFIGURATION_DIRECTORY", "/etc/arm/config")
```
microtechno9000: Where is this defined? I.e. how is the system pulling the environment variable?

arm/config/config.py
```diff
@@ -28,7 +28,13 @@ def _load_abcde(fp):
 # handle arm.yaml migration here
 # 1. Load both current and template arm.yaml
 cur_cfg = _load_config(arm_config_path)
-new_cfg = _load_config("/opt/arm/setup/arm.yaml")
+new_cfg = _load_config(os.path.join(cur_cfg['INSTALLPATH'], "setup/arm.yaml"))
```
microtechno9000, line 35: Similar comment, where are these values defined?

microtechno9000 (changes requested):
Good idea and will help make changes easier, as some users have been keen to not run ARM from /opt.
However, I dont see where the environment variables are being defined. Is there a .env file thats missing, or maybe I miss understand what's happening here.

arm/config/config.py
xieve: Setting this env var is optional, the default value if unset is `"/etc/arm/config"`. With my systemd service, it's set by systemd to the value of the `ConfigurationDirectory` parameter. You can read about them here: [systemd.exec(5)](https://man.archlinux.org/man/core/systemd/systemd.exec.5.en#SANDBOXING) (scroll down a bit to the table, it lists the config options and what env var they translate to)

arm/config/config.py
xieve: This one is defined in arm.yaml, in the case of the NixOS module I override that setting in my module with the correct path, this was not included in the systemd example I posted above.

arm/config/config.py
microtechno9000: Can you update the wiki to provide examples of how to set the .env variables required to modify / override the config

xieve: As I don't have maintainer access, I'm not able to edit the wiki afaict, but here's my proposition:

Under [Configuring ARM](https://github.com/automatic-ripping-machine/automatic-ripping-machine/wiki/Configuring-ARM):

<details>
<summary>Markdown to copy-and-paste</summary>

    # Environment variables
    
    > [!WARNING]
    > This is an advanced topic and only relevant to you if you want to set up ARM in a non-standard way. It is especially irrelevant when using docker.
    
    ARM (or rather, the ARM Python modules for ripper and UI) optionally accept a number of [environment variables](https://wiki.archlinux.org/title/Environment_variables) to override certain paths, to provide more flexibility concerning the environment you run it in. These always override any settings set in `arm.yaml`.
    
    | Environment variable | Contents | Default value | Related setting in `arm.yaml` |
    | --- | --- | --- | --- |
    | `CONFIGURATION_DIRECTORY` | Config files (`arm.yaml` etc.) | `/etc/arm/config` | N/A |
    | `RUNTIME_DIRECTORY` | Mountpoints for disks | `/mnt` | N/A |
    | `STATE_DIRECTORY` | Database | `/home/arm/db` | `DBFILE`¹ |
    | `LOGS_DIRECTORY` | Logs | `/home/arm/logs` | `LOGPATH` |
    
    ¹`DBFILE` (in `arm.yaml`) specifies the exact file path of the DB file, while `STATE_DIRECTORY` only specifies the directory in which the database, which is then always called `arm.db`, is saved.
    
    The variable names are selected such that they are compatible with certain Systemd parameters, to facilitate running ARM as a service. You can read more about them on the [systemd.exec(5)](https://man.archlinux.org/man/core/systemd/systemd.exec.5.en#SANDBOXING) manpage.
    
    ## Example
    
    _How_ you set environment variables depends on your specific way of invoking ARM (e.g. init script, Systemd service…), but here is an example for a simple wrapper script:
    
    ```bash
    #!/usr/bin/env bash
    # my-armui-wrapper.sh
    
    export CONFIGURATION_DIRECTORY="/etc/arm"
    export RUNTIME_DIRECTORY="/run/arm"
    
    # $@ means forward all arguments
    /path/to/armui $@
    ```
    
</details>

# Environment variables

> [!WARNING]
> This is an advanced topic and only relevant to you if you want to set up ARM in a non-standard way. It is especially irrelevant when using docker.

ARM (or rather, the ARM Python modules for ripper and UI) optionally accept a number of [environment variables](https://wiki.archlinux.org/title/Environment_variables) to override certain paths, to provide more flexibility concerning the environment you run it in. These always override any settings set in `arm.yaml`.

| Environment variable | Contents | Default value | Related setting in `arm.yaml` |
| --- | --- | --- | --- |
| `CONFIGURATION_DIRECTORY` | Config files (`arm.yaml` etc.) | `/etc/arm/config` | N/A |
| `RUNTIME_DIRECTORY` | Mountpoints for disks | `/mnt` | N/A |
| `STATE_DIRECTORY` | Database | `/home/arm/db` | `DBFILE`¹ |
| `LOGS_DIRECTORY` | Logs | `/home/arm/logs` | `LOGPATH` |

¹`DBFILE` (in `arm.yaml`) specifies the exact file path of the DB file, while `STATE_DIRECTORY` only specifies the directory in which the database, which is then always called `arm.db`, is saved.

The variable names are selected such that they are compatible with certain Systemd parameters, to facilitate running ARM as a service. You can read more about them on the [systemd.exec(5)](https://man.archlinux.org/man/core/systemd/systemd.exec.5.en#SANDBOXING) manpage.

## Example

_How_ you set environment variables depends on your specific way of invoking ARM (e.g. init script, Systemd service…), but here is an example for a simple wrapper script:

```bash
#!/usr/bin/env bash
# my-armui-wrapper.sh

export CONFIGURATION_DIRECTORY="/etc/arm"
export RUNTIME_DIRECTORY="/run/arm"

# $@ means forward all arguments
/path/to/armui $@
```

xieve: I figured out what I was supposed to do :)

xieve:
Okay, basically rewrote the PR, imo it is much clearer now what I'm doing. Instead of five env vars, I'm now using just `ARM_CONFIG_FILE`, and all other parameters are read from the config. I need to make some finishing touches now though.

xieve:
I would like to use `mount --all /dev/sr<number>` to mount the drives, but all the documentation on manual installs instructs users to add these mount options to their `fstab`:
```
users,noauto,exec,utf8,ro
```
Which strikes me as odd twice:
1. Why `noauto`? (It inhibits the use of `mount --all <devpath>`)
2. Why `exec`? AFAIK ARM never wants to run binaries off of optical media

Here's what I would suggest (maybe I'll open another PR):
```
defaults,users,utf8,ro
```

Anyway, while this would be easy to change for docker users, the setups of people with a manual install would likely break. I deemed this *probably bad* and worked around it by querying `fstab` via `findmnt`, but IMO the resulting code is much more convoluted and unclear than it needs to be. LMK if you would be fine with this breaking change.

Edit: testing is now gone through, here's the new log:

[2_MULES_FOR_SISTER_SARAH_176298319677.log](https://github.com/user-attachments/files/23510764/2_MULES_FOR_SISTER_SARAH_176298319677.log)

Landing this PR would now be fine for me, if you don't want me to make the change I outlined above.

arm/ripper/main.py
```diff
@@ -13,11 +13,15 @@
...
 import getpass  # noqa E402
+from importlib.util import find_spec
+from pathlib import Path
 import pyudev  # noqa: E402
 import psutil  # noqa E402
 
-# set the PATH to /opt/arm so we can handle imports properly
-sys.path.append("/opt/arm")
+# If the arm module can't be found, add the folder this file is in to PYTHONPATH
+# This is a bad workaround for non-existent packaging
```
microtechno9000, line 22: Why is the package nonexistent? Is this a check on if the root 'arm' module has been pulled in?

microtechno9000 (changes requested):
some minor comments

microtechno9000:
> I would like to use `mount --all /dev/sr<number>` to mount the drives, but all the documentation on manual installs instructs users to add these mount options to their `fstab`:
> 
> ```
> users,noauto,exec,utf8,ro
> ```
> 
> Which strikes me as odd twice:
> 
>     1. Why `noauto`? (It inhibits the use of `mount --all <devpath>`)
> 
>     2. Why `exec`? AFAIK ARM never wants to run binaries off of optical media
> 
> 
> Here's what I would suggest (maybe I'll open another PR):
> 
> ```
> defaults,users,utf8,ro
> ```
> 
> Anyway, while this would be easy to change for docker users, the setups of people with a manual install would likely break. I deemed this _probably bad_ and worked around it by querying `fstab` via `findmnt`, but IMO the resulting code is much more convoluted and unclear than it needs to be. LMK if you would be fine with this breaking change.
> 
> Edit: testing is now gone through, here's the new log:
> 
> [2_MULES_FOR_SISTER_SARAH_176298319677.log](https://github.com/user-attachments/files/23510764/2_MULES_FOR_SISTER_SARAH_176298319677.log)
> 
> Landing this PR would now be fine for me, if you don't want me to make the change I outlined above.

The existing fstab rules are possibly a hang over from when ARM was not installed and maintained within docker. If a user installed arm on a system with Gnome or KDE, then the automount would mount the drive within the OS.
Noting that docker is the main supported installation method, adjusting to suit is fine. A warning can be added to the manual user wiki install guides that modification to their fstab would be required for ARM to work.

arm/ripper/main.py
xieve: What I mean is, ARM should be packaged as a Python package (I've already done that on my fork). That would make this hack unnecessary, technically.

Edit: Sort of, `find_spec` checks if the `arm` module can be found in `sys.path` already (e.g. because it has been installed as a setuptools package, such as in my NixOS setup).

xieve:
Alright! I'll make the change right now, but I'm not home, so it might take until monday before I can test that.

xieve:
Here's the interdiff for the latest change:
https://github.com/automatic-ripping-machine/automatic-ripping-machine/compare/09f0283f610e525c1a9132a6b17d0243570b2c42..1e877f23e1110fc4e548c7e195211f124c6f1d03

IMO it's much simpler now. Should I make it clear on the alternate install method pages that there has been a change to the `fstab`, or is it fine as it is?

xieve:
Upon closer inspection, it seems I have introduced an error into the title identification. I am not sure how yet.

microtechno9000 (approved):

PR #1507, opened 2025-08-10
[BUGFIX] abandon job with non existing pid

### LINKED ISSUE #1506: 🐛 Cannot release Job after server restart
### Is there an existing issue for this?

- [x] I have searched the existing issues

### I have checked the wiki

- [x] I have searched the wiki for this problem

### Does this issue exist in the latest version?

### Describe the bug?

After a server restart, unfinished ripping processes remain visible in the UI and cannot get abandoned. The only option is to delete the job from the database.

The root cause behind this is that the database still holds the old job id but since the server restartet, this process id is not valid anymore. Non existing Process ids can safely be assumed terminated:

https://github.com/automatic-ripping-machine/automatic-ripping-machine/blob/f74d89af0371d9a8bfab81557184ecdc97e860f3/arm/ui/json_api.py#L434-L437

### To Reproduce

Start Job
Restart Server
Press Abandon in UI

### What OS are you seeing the problem on?

Other (please state in Anything else? section)

### ARM Release Version or Git hash (You can find this in the arm settings page)

Ubuntu 22.04, f74d89af0371d9a8bfab81557184ecdc97e860f3 2.18.1

### Expected behavior?

Abandoning Jobs should work for otherwise terminated processes

### Relevant log output

```shell
[2025-08-10 14:08:25,325] DEBUG ARM: json_api.abandon_job Couldn't find job.pid - 2121141! Reverting db changes.
```

### Anything else?

_No response_

### Code of Conduct

ukos-git:
# Description

Handle Non-Existing PIDs while releasing job

Fixes #1506

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?
Ubuntu 22.04, bare metal, release running jobs, release non-existing jobs.

# Checklist:

- [X] I have commented my code, particularly in hard-to-understand areas

# Changelog:

This wraps the psutil functions into its own try-catch loop. After the
psutil loop, a process is either terminated or an error occurred that is
not yet handled.
If the process is terminated, we return "true". If an error occurred, we
revert the database so that the job will still be visible and the user
can try to kill it once more.

The prerequisite for releasing the drive from the current job is that the
associated process is terminated. Otherwise the drive is not released.
Releasing the drive from the job is completely handled by job.eject().
No call to a (possibly non-existing) associated drive is done on the
abandon-layer.

# Logs
before:
```
[2025-08-10 14:08:25,325] DEBUG ARM: json_api.abandon_job Couldn't find job.pid - 2121141! Reverting db changes.
```

after (job with unknown PID)
```
[2025-08-10 16:25:50,153] DEBUG ARM: json_api.get_x_jobs jobs  - we have 1 jobs
[2025-08-10 16:25:51,497] WARNING ARM: json_api.abandon_job Process id 2121141 was not found. Job has already been terminated.
[2025-08-10 16:25:51,497] DEBUG ARM: job.eject The drive associated with this job has already been ejected.
...

after (job with known PID)
...
[2025-08-10 17:01:08,914] DEBUG ARM: json_api.abandon_job Job 934 was terminated.
...

### DIFF
--- a/arm/ui/json_api.py
+++ b/arm/ui/json_api.py
@@ -416,43 +416,52 @@ def abandon_job(job_id):
         return json_return
 
     # Kill the process id
+    job = Job.query.get(int(job_id))
+    job.status = JobState.FAILURE.value
     try:
-        job = Job.query.get(int(job_id))
-        job.status = JobState.FAILURE.value
-        try:
-            job_process = psutil.Process(job.pid)
-            job_process.terminate()  # or p.kill()
-        except psutil.NoSuchProcess:
-            message = f"Process id {job.pid} was not found. Job has already been terminated."
-            app.logger.warning(message)
-            # No raise here. No process is a terminated process.
-        except psutil.AccessDenied as err:
-            message = f"Access denied abandoning job: {job.pid}!"
-            app.logger.error(message)
-            raise ValueError(message) from err
-        else:
-            app.logger.debug(f"Job {job_id} was terminated.")
-    except Exception as error:
-        json_return["Error"] = str(error)
+        terminate_process(job.pid)
+    except Exception as err:
+        db.session.rollback()
+        json_return["Error"] = str(err)
         json_return['success'] = False
+        title = f"Job ERROR: {job.pid} couldn't be abandoned."
+        message = json_return['Error']
+        app.logger.debug(f"{title} - Reverting db changes - {message}")
+        notification = Notifications(title, message)
     else:
         job.eject()  # only release/eject the job if the process got killed.
         json_return['success'] = True
-
-    # Handle Errors
-    if 'Error' in json_return:
-        title = f"Job ERROR: {job_id} couldn't be abandoned."
-        message = json_return['Error']
-        app.logger.debug(f"{title} - Reverting db changes - {message}")
-        db.session.rollback()
+        title = f"Job: {job.pid} was Abandoned!"
+        message = f'Job with id: {job.pid} was successfully abandoned. No files were deleted!'
         notification = Notifications(title, message)
-    else:
-        notification = Notifications(
-            f"Job: {job_id} was Abandoned!",
-            f'Job with id: {job_id} was successfully abandoned. No files were deleted!')
     db.session.add(notification)
     db.session.commit()
     return json_return
+
+
+def terminate_process(pid):
+    """
+    Terminates the process associated with a given pid.
+    :param pid: Process ID (int)
+    :raises: ValueError if access is denied
+    """
+    if pid is None:
+        message = "PID not found for job."
+        app.logger.warning(message)
+        return
+    try:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: arm/ui/json_api.py [hunk 1]
```diff
@@ -415,38 +415,43 @@ def abandon_job(job_id):
...
+    # Handle Errors
     if 'Error' in json_return:
-        notification = Notifications(f"Job ERROR: {job_id} couldn't be abandoned", json_return["Error"])
-        db.session.add(notification)
-        db.session.commit()
+        title = f"Job ERROR: {job_id} couldn't be abandoned."
+        message = json_return['Error']
+        app.logger.debug(f"{title} - Reverting db changes - {message}")
+        db.session.rollback()
+        notification = Notifications(title, message)
```
microtechno9000, line 430: this notification isn't committed to the db

arm/ui/json_api.py
```diff
@@ -415,38 +415,43 @@ def abandon_job(job_id):
...
+    else:
+        job.eject()  # only release/eject the job if the process got killed.
+        json_return['success'] = True
+
+    # Handle Errors
     if 'Error' in json_return:
-        notification = Notifications(f"Job ERROR: {job_id} couldn't be abandoned", json_return["Error"])
-        db.session.add(notification)
-        db.session.commit()
+        title = f"Job ERROR: {job_id} couldn't be abandoned."
```
microtechno9000: I don't see where in the code that the job is abounded and written to the database. if there is in running pid, the job should still be updated to 'failed'to prevent it reporting as active on the webpage

VERSION
```diff
@@ -1 +1 @@
-2.18.1
+2.18.2
```
microtechno9000: bump to .3

microtechno9000 (changes requested):

arm/ui/json_api.py
ukos-git: i add both notifications after the conditional branch.

arm/ui/json_api.py
ukos-git: Thanks for the input!

Changes to the jobs are only updated if the process could get terminated. Abandoning is not possible if the PID is active.

I prefer not marking jobs as failed while the PID is still running, so the user can retry or intervene. Once the PID disappears, the job can be properly abandoned but I don't like to have active PIDs (possibly blocking a drive) that are invisible.

Let me know what you think.

ukos-git: I restructured the job termination into another function. This way it looks way cleaner. Tests were perfomed on jobs on "waiting" state with running PID. Abandoned PIDs "should" work, though not tested again.

microtechno9000: Thanks, I will have a look at the changes. Abandoning jobs has been a bug bear for a while, either failing or leaving hanging PIDs

PR #1514, opened 2025-08-11
[BUGFIX] Verifying checksum of downloaded source Tarball

dbstf:
# Description
Due to a copy-paste error, the SHA256 checksum of the downloaded archive for the binary `makemkv-bin-...` gets checked twice, whereas the downloaded archive for the source `makemkv-oss-...` does not get checked at all.

## Type of change

# How Has This Been Tested?
Running the installer script and corresponding commands manually in the terminal.

- [x] Other (Debian 12 direct install)

# Checklist:

# Changelog:
Changed L612 from
`grep "makemkv-bin-${LatestMakeMKVVersion}.tar.gz" "makemkv-sha-${LatestMakeMKVVersion}.txt" | sha256sum -c`
to
`grep "makemkv-oss-${LatestMakeMKVVersion}.tar.gz" "makemkv-sha-${LatestMakeMKVVersion}.txt" | sha256sum -c`


VERSION
```diff
@@ -1 +1 @@
-2.18.2
```
microtechno9000: please update to align with current version

microtechno9000 (changes requested):
Please bump revision from 2.18.3 to 2.18.4

PR #1542, opened 2025-08-30
[FEATURE] Use docker compose as an alternative to bash file

MartinFixes:
Rearrangement of the "start_arm_container.sh" file into a "docker-compose.yaml" file that can be run using the command "docker compose up -d"

The benefit of this pull request is that it makes running and stopping the software more straightforward and standard.


microtechno9000:
@MartinFixes thanks for this change, agree docker compose is much easier.

Please add some information to the wiki to help new users find the docker compose file and how to edit it.

Noting that there is a difference between docker, and docker compose with additional packages required to use docker compose depending on your base system.

Help can be found here https://github.com/automatic-ripping-machine/automatic-ripping-machine/wiki/Contribute-Wiki

microtechno9000 (approved):
Nice, simple addition, thanks

microtechno9000 (approved):

microtechno9000 (changes requested):
changes required to the wiki documentation

PR #1554, opened 2025-09-07
[BUGFIX] make "scan for drives" look like a button

kpupp:
this simply added the .btn and related classes to the Scan for Drives link to make it look properly actionable.
Screenshot 2025-09-07 at 2 21 12 PM


kpupp:
it seems this may need changing somewhere else - is the HTML not static?

microtechno9000:
Is this PR ready for incorporation?

kpupp:
@microtechno9000 I'm not sure - I never got clarification on how that button is made. When I made these changes (adding the class) it wouldn't stick so I think it's generated by some other code somewhere in the program.

microtechno9000:
> @microtechno9000 I'm not sure - I never got clarification on how that button is made. When I made these changes (adding the class) it wouldn't stick so I think it's generated by some other code somewhere in the program.

That is the only spot that the 'Scan for Drives' is called from within ARM.

Testing locally, adding in the below code works in both dark and light mode, I used 'btn-sm' not large

```python
            <p><a class="btn btn-secondary btn-sm" href="systemdrivescan">Scan for Drives</a></p>
```

image

image

kpupp:
Got it - I'll give it another whack, unless that works well then just use
that. :)

microtechno9000:
Closing as stale

PR #1565, opened 2025-09-23
[FEATURE] Update memory together with CPU

martinnj:
# Description
Currently the 'total memory' of the system is only detected on the very first launch, then written to the database and never saved again.

This PR changes this so that when the `/updatecpu` function is called it also refreshes the total memory of the system.

I debated on whether or not this was a bugfix or feature, since the memory being wrong could be considered a bug, but it's also not like it crashes or really changes anything, so I settled on feature.

## Type of change

# How Has This Been Tested?

Changed where copied into a freshly pulled container and tested there.

# Checklist:

# Changelog:

- "Update CPU" button in settings now also updates system memory totals.

# Logs


1337-server:
Pulled and tested, no issues.

However, it might be better if the `Update CPU` was changed to include the memory, so it's clear it refreshes both.

martinnj:
> However, it might be better if the `Update CPU` was changed to include the memory, so it's clear it refreshes both.

Makes perfect sense. I've renamed it, I decided to just rename it after the class instead, that way if the `SystemInfo`class gets updated to detect proper hostnames, etc, the frontend doesn't actually need to change. :)

1337-server:
Thanks for your contribution to ARM!

PR #1580, opened 2025-10-02
UI safety updates rebranch

flammableliquids:
# Description

The forms "Ripper Settings" And "UI Settings" Have now been upgraded.
This includes the beginnings of field validatiors and drop downs.
"Ripper Settings" form has too many fields to be defined manually in python, and so a json has been created to allow for ease of organization and management of the form. 

Fixes # (issue title here)

# How Has This Been Tested?
Many many repeated tests in 'bare metal mode'

- [ ] Docker
- [x] Other (24.04 Ubuntu)

# Checklist:

# Changelog:

Most recent fixes and improvements:
- write to arm.yaml logging has been improved.
- - better logging
- - cleaner data-typing for json to yaml
- commented debug messages have been removed.
- custom validators for dynamically created forms work best when the validators are classes not functions.
- dynamic validator creation has been moved into functions outside of class creation, making code easier to read.
- testing successful.

**I am still trying to figure out how to modify the jinja to show the validation errors.**
**It will now prevent people from saving dangerous or wrong configurations, by erroring to a log file,**
**_and giving no visual feedback to user._**
(I'm still trying to figure out how to format the form in jinja macros to have the error come up, and the users focus re-directed.
I felt it best to make the pr to prevent people from screwing up)

- arm/ui/settings.SettingsForm() is not hard-coded for rendering, if a new field is added to the form, the loop will now render it.
- applied to "UI Settings":
- - DataRequired validator now 
- - Hinting now applied to fields (not just the blue 'i' icon but on hover.
- - use_icons and save_remote_images changed to dropdowns to prevent user mistakes.
- arm/ui/forms.SettngsForm() completely re-created:
- - is no longer a class, but a function that returns an instance of a class.
- - creates an instance an empty class, and then based on ripperFormConfig.json populates the fields(types)
- - Hinting now applied to fields.
- - Dropdown fields and integer fields now enforce user input to prevent users from crashing their servers (This all started because I typed DEBRUG instead of DEBUG and had to plug in to the server to find the mistake). 
- ripperFormConfig.json created, and quickJsonGenerator.py (to create ripperFormConfig.json), so that the field types, and validators for the "Ripper Settings" form can be easily managed via loops instead of massive python code.
- arm/ui/forms_custom_validators.py added. To hold custom validators. Primary goal validator is the "validate_path_exists" to prevent users from inputting files or folders that do not exist on the server. (If you do this, you will cause your server to crash) (validator either does not work, or the rendering of form error fields doesn't show the error)
- improved datatyping in forms.py
- arm/ui/utils.py Updated: Improved dataype indications and function return type indications. Functions modified or added:
- - listCoPairedIntoTuple (Ability in name).
- - jsonFile_to_dict : provided a str|Path will read the file and return a dict.
- - generate_comments modified to use jsonFile_to_dict
- - generate_ripperFormSettings reads ripperSEttingsConfigFile.json and uses jsonFile_to_dict to return a dict
- Jinja Changes:
- - _form_field.html created, containing the marco 'render_field(field)', takes a field, and renders it, including the 'i' popover.
- - arm/ui/settings/templates/settings/ui.html imports render_field,  and instead of manually executing against each field in the form, loops over the passed form, and calls render_feild
- - arm/ui/settings/templates/settings/ripper.html imports render_field,  and instead of manually executing against each field in the form, loops over the passed form, and calls render_feild
- [BugFix] arm/runui.py updated for better path support.
- [BugFix] arm/ui/comments.json was Missing the UMASK Field
- README Fix Including link correction
- dynamic form creation has been made more modular. Making it easier to read, and reusable for when other forms are made dynamic.
- extraneous comments have been removed

SonarCubeCloud will generate a security warning, because I am using Jinja to render fields in HTML, considering that this project is not meant for 'production web' implementations, I am willing to accept the unlikely security risk


arm/runui.py
```diff
@@ -4,7 +4,8 @@
 import sys
 
 # set the PATH to /arm/arm, so we can handle imports properly
-sys.path.append(os.path.join(os.path.dirname(os.path.abspath(__file__)), ".."))
+# sys.path.append(os.path.join(os.path.dirname(os.path.abspath(__file__)), ".."))
```
1337-server: Remove the commented out code

arm/ui/settings/settings.py
```diff
@@ -176,27 +188,43 @@ def save_settings():
...
+        app.logger.debug(f"Cleansed Form Data: \r\n{arm_cfg}")
+        try:
+            # Save updated arm.yaml
+            app.logger.debug(f"routes.save_settings: Saving new arm.yaml: {cfg.arm_config_path}")
+            with open(cfg.arm_config_path, "w") as settings_file:
+                settings_file.write(arm_cfg)
+            success = True
+        except Exception as e:
+            app.logger.exception(f"Error saving arm.yaml: {e}")
+            flash(f"Error saving arm.yaml: {e}", "error")
```
1337-server, line 215: I thought we moved away from flash() ? And were using toasts/notification system ?

Because the flash() doesn't update till the next page load, so there will be no user feedback until they go to another page. You can see this, by failing to update the settings and then changing the page, a badly formed box will appear. But only after then user has already changed the page.

arm/ui/settings/settings.py
```diff
@@ -176,27 +188,43 @@ def save_settings():
...
         # Set the ARM Log level to the config
         app.logger.info(f"Setting log level to: {cfg.arm_config['LOGLEVEL']}")
         app.logger.setLevel(cfg.arm_config['LOGLEVEL'])
 
     # If we get to here there was no post data
-    return {'success': success, 'settings': cfg.arm_config, 'form': 'arm ripper settings'}
+        return {'success': success, 'settings': cfg.arm_config, 'form': form_name}
+    else:
+        app.logger.error(f"Error validating form: {form.errors}")
+        flash(f"Error validating form: {form.errors}", "error")
```
1337-server, line 226: See previous, comment.

arm/ui/settings/templates/settings/ui.html
```diff
@@ -3,113 +3,14 @@
...
-            </div>
+            {% from "settings/_form_field.html" import render_field %}
+            {% for field in ui_settings %}
+                {% if field.name != 'submit' %}
+                    {% if field.name != 'csrf_token' %}
+                    {{ render_field(field) }}
+                    {% endif %}
+                {% endif %}
+            {% endfor %}
             <button id="uiSubmit" class="btn btn-secondary btn-lg btn-block" form="uiSettings"
```
1337-server, line 14: This no longer submits when the return/enter key is pressed.

Also, the layout is broken and will need the CSS looked at, as all the inputs are not correctly evenly sized

arm/ui/templates/base.html
```diff
@@ -10,9 +10,7 @@
             {% with messages = get_flashed_messages(with_categories=true) %}
                 {% if messages %}
                     {% for category, message in messages %}
-                        <div id="message1" class="alert alert-{{ category if category !=" message
-            " else 'info' }} text-center"
-                             role="alert">
+                        <div id="message1" class="alert alert-{{ category if category != "message" else 'info' }} text-center" role="alert">
```
1337-server, line 13: See comment about flash. These are no longer used as it's better to use the notification system.

1337-server (changes requested):
Still some work to do. I would really advise against using the flash() as this doesn't work well with everything else updating instantly

Return/Enter key should still submit forms.

But the changes to make true/false much simpler is great.

arm/ui/settings/templates/settings/ui.html
flammableliquids: The functionality to save on enter has not been removed, it's the main method I have been using to test this (CBF scrolling all the way down).

You might find that the arm.yaml your testing against is throwing form based validation errors. 
settings_form_error

Those are form validation errors, for the 'template' arm.yaml.

arm/ui/settings/settings.py
flammableliquids: I'm not sure if we moved away from it in the past.
I was looking into handling the 'Flash' via ajax.

Your correct throwing the error to the user on form submit is best, waiting for page refresh is ridiculous.
But until I can sort out the ajax end, it's best to leave these in as a sort of 'placeholder', IMHO.

flammableliquids:
Based on your testing, and additional re-runs of testing, the validators are flagging problems, which then prevents 'hitting enter' to save. 
I will try and figure out what part broke 'enter to save'

arm/ui/templates/base.html
flammableliquids: This change was to remove a carriage return that might be the cause of problems when rendering the page.

arm/ui/settings/settings.py
microtechno9000: We have moved some notifications from flash to the toast/notification calls, but not all

arm/ui/settings/templates/settings/ui.html
microtechno9000: Is there an issue with the template or the new parsing of the arm.yaml template/default?

microtechno9000:
@flammableliquids suggest joining the discussion on Version 3.0 on discord, a lot of the changes that you have added will be resolved with 3.0
Whilst you have made a lot of great changes here; however, once we move to 3.0 all of it will be lost. Would love to get your input to ensure we don't run into the same issue with the new version

PR #1604, opened 2025-10-19
[FEATURE] Changed build base image 

1337-server:
# Change base image of arm-ui devel image

# Description
Changed build base image, as we don't need such a large image if we are only running one service (frontend)
     - **Image shrunk from 1GB -> 160MB**

Removed comments from Dockerfile
    - **Way to many (IMHO) and was harder to read**

Broke UI requirements to its own file, to help track requirements separately
    - **It will be easier to track, and help keeps things manageable to have separate requirements for each container**

Removed comments from health check, also udev isnt needed in UI container so was removed.
    - **Comments didnt really help, as for the ripper we know already what needs checking for the container to work, so the comments dont help**

Fixes # None

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?
Build, ran on docker.

- [x] Docker

# Checklist:

# Changelog:

Summary of changes

   - Dockerfile-UI
       - Rewritten to use python:3.12.1-alpine with a builder stage to compile/install Python deps and a minimal final runtime stage.
       - Sets ARM_UID/ARM_GID to 1000, PYTHONUNBUFFERED and PIP_NO_CACHE_DIR.
       - Installs build deps in the builder and copies installed packages into the final image.
       - Creates non-root arm user (uid/gid 1000) and runs the container as USER arm.
       - Copies repo into /opt/arm and sets git safe.directory.
       - Removes Debian/my_init/runit and ripper-specific install steps from the UI image.
       - Default CMD changed to run the UI directly: python3 /opt/arm/arm/runui.py
    requirements_ui.txt (new)
       - New, UI-focused dependency list (Flask 3.x, Werkzeug, Jinja2, SQLAlchemy + Alembic/Flask-Migrate, mysql-connector-python, waitress, psutil, pyudev, Flask-Login, Flask-WTF, bcrypt, etc.).
        Helps separate UI deps from ripper/tooling dependencies.
      - Added a copy `comments.json` as arm now looks for it inside `/arm/app/comments.json`

  -  docker-compose.yml
       - ARM_UID and ARM_GID environment variables updated from 1001 -> 1000.
       - arm-db user mapping updated to 1000:1000.
 -   temp_healthcheck.sh
       - Simplified to only perform an HTTP check against the UI endpoint (removed udev and ripper checks).

# Logs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -161,6 +161,19 @@ RUN install_clean pip curl git libcurl4-openssl-dev \
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/doc/* /usr/share/man/* /usr/share/locales/* \
     && rm -rf /etc/service/sshd /etc/my_init.d/00_regen_ssh_host_keys.sh
 
+RUN chmod +x /etc/my_init.d/*.sh && \
+    ldconfig && \
+    strip HandBrakeCLI \
+    && strip makemkvcon || echo "makemkvcon not found yet"
+
+# Add dvd encryption support - not always needed but helpfull if user only uses HandBrakeCLI for dvds
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y libdvd-pkg && \
+    echo "libdvd-pkg/libdvd-pkg/auto-install boolean true" | debconf-set-selections && \
+    dpkg-reconfigure -f noninteractive libdvd-pkg && \
+    rm -rf /var/lib/apt/lists/*
+
+
 # Add ARMui service
 RUN mkdir /etc/service/armui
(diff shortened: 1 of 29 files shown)


1337-server:
Currently, we are using `phusion/baseimage` as a base for the front end, but there is no need to use this for just the UI
The size also bloats massively as we include all the ripper dependencies from pip

The current docker also uses `automaticrippingmachine/arm-dependencies:1.1.8` during building for the front end, this again adds massive bloat to the final image, especially if we are running the ripper in its own container.

This removes the bloat and leaves the UI completely independent

microtechno9000:
Propose closing this as #1754 is further along in the development of v3.x and will be easier to update to a newer version of phusion image.

PR #1652, opened 2025-12-24
 [FEATURE] Docker Install Script: add support for more Distros. Change red text to green text for better clarity

53845714nF:
# Description
I add some more Distros in the docker install script. Motivation is that Docker is independent of the operating system. More and more people are using Arch based Linux. For example, Cachy OS or Manjaro.

## Type of change
- [x ] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# How Has This Been Tested?

- [ ] Docker
- [x] Other (Cachy OS)
- [x] Shellcheck

# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas

# Changelog:

- The Docker script can be used on various distributions. 
- Change red text to green text for better clarity.


microtechno9000:
@53845714nF not sure I agree with this change. The reason for moving ARM to a known Debian based image is that when it builds and runs we (the developers) know what is in the container and thus can limit the unknown unknowns with free and open software.

The base ARM image is built using phusion/baseimage:jammy-1.0.4
https://github.com/automatic-ripping-machine/arm-dependencies/blob/aeffc5400263163f2c09c759fb4b8e92ac4c3e6d/Dockerfile#L3

I don't see where adding in a step here for other distro's helps. The docker container is agnostic of the host Linux variant.

Before I close this request, can you please provide some additional information or context if I am missing it.

53845714nF:
The PR does not affect the Dockerfile. It ensures that the Docker Engine is installed on the machine so that the container can be started.

I would find that helpful, as more and more people are using Manjaro or cachey os.

mihawk90:
I don't have any say in whether this gets accepted, but IMO an install script is not the place for this. I don't think it is too much to ask of a user wanting to install a Docker container to read the Docker documentation on how to actually get Docker to work in the first place. The ARM documentation already links to the correct Docker docs page to install it, so by putting it into the install script it makes one of the two redundant.

An install script (if we really have to have one) should take care of installing *that specific piece of software*, not everything possibly related to it. That is the job of whoever is wanting to use said piece of software.

---

On a side tangent:

The install script is IMO already a crutch because we have a bunch of users that run it without actually knowing what they're doing, and then they come here or to the Discord with basic Docker issues that aren't even directly related to ARM.
The majority of projects providing a Docker container simply have an example compose file and let the user figure it out from there, because compose is well documented already and the assumption is that the user has the prerequisite knowledge to install packages and assign users to a group.

53845714nF:
The script was my first point of contact with automatic-ripping-machine. 
I just wanted to make it more convenient for users who might not have Debian-based Linux.
I think just a Compose file would also be good; there is Pull Request #1542, which I would also very much welcome.

microtechno9000:
I see, I miss read what this script is doing. I haven't used these scripts and would rather not support/maintain them due to the ever changing nature of dependencies.

With regards to using docker compose there is nothing stopping you creating a docker compose file and installing ARM. Something on my list to add to ARM, just haven't.

I am happy to merge this file, if you can 
1. Edit it and add a note that this script is not maintained by ARM developers and may not function as expected (something along those lines)
2. Update the wiki to include information on the script and the various distros you are adding, with the above warning/note

Thanks though, for adding to ARM and supporting the project :)

microtechno9000 (changes requested):
Wiki update required to https://github.com/automatic-ripping-machine/automatic-ripping-machine/wiki/docker
Code is here in the repo https://github.com/automatic-ripping-machine/automatic-ripping-machine/blob/main/arm_wiki/Docker.md

Please update to include details for other distros now included

53845714nF:
I added that to the wiki.

scripts/installers/docker-setup.sh
```diff
@@ -26,51 +34,90 @@ done
...
+            apt update -y && apt upgrade -y
+            apt install -y curl lsscsi
+            ;;
+        arch|cachyos|manjaro)
+            pacman -Syu --noconfirm curl lsscsi
+            ;;
+        fedora)
+            dnf install -y curl lsscsi
+            ;;
+        rhel|centos|rocky|almalinux)
```
mihawk90: TLDR use `dnf` for RHEL

---

Unless we really want to support RHEL 7 - which is in the final stages of dying to EOL - this block should be unified with the Fedora block.

`yum` has been deprecated in RHEL since RHEL 8 in 2019(!) and Red Hat recommends using DNF instead:
https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html-single/managing_software_with_the_dnf_tool/index#con_software-management-tools-in-red-hat-enterprise-linux-9_managing-software-with-the-dnf-tool

The [RHEL 10 documentation](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/10/html-single/managing_software_with_the_dnf_tool/index) on the other hand makes no mention of yum at all (except for the DNF config directories for historical reasons). I don't actually know if `yum` is still symlinked to `dnf` in RHEL 10, but if it is not, then this will immediately break the script.

scripts/installers/docker-setup.sh
```diff
@@ -26,51 +34,90 @@ done
...
         passwd arm
     else
-        echo -e "${RED}arm user already exists, skipping...${NC}"
+        echo -e "${GREEN}arm user already exists, skipping...${NC}"
     fi
-    usermod -aG cdrom,video arm
+    
+    getent group cdrom >/dev/null || groupadd cdrom
+    getent group video >/dev/null || groupadd video
+    usermod -aG cdrom,video arm && echo -e "${GREEN}Adds user arm to cdrom, video group${NC}"
```
mihawk90: Should be "Added". "Adds" implies it's about to do it, but this being after a `&&` means it is already done.

I would also add quotation marks to the user- and groupnames, i.e. "arm", "cdrom".

The `groupadd` should probably also get a feedback like this.

scripts/installers/docker-setup.sh
```diff
@@ -26,51 +34,90 @@ done
...
 }
 
 function launch_setup() {
     # install docker
     if [ -e /usr/bin/docker ]; then
-        echo -e "${RED}Docker installation detected, skipping...${NC}"
-        echo -e "${RED}Adding user arm to docker user group${NC}"
-        usermod -aG docker arm
+        echo -e "${GREEN}Docker installation detected, skipping...${NC}"
+        usermod -aG docker arm && echo -e "${GREEN}Adds user arm to docker user group${NC}"
```
mihawk90: Same as above.

scripts/installers/docker-setup.sh
```diff
@@ -26,51 +34,90 @@ done
...
-        echo -e "${RED}Installing Docker${NC}"
+        echo -e "${GREEN}Installing Docker${NC}"
         # the convenience script auto-detects OS and handles install accordingly
-        curl -sSL https://get.docker.com | bash
-        echo -e "${RED}Adding user arm to docker user group${NC}"
-        usermod -aG docker arm
+        case "$OS_ID" in
+            debian|ubuntu|fedora|rhel|centos|rocky|almalinux)
+                curl -sSL https://get.docker.com | bash
+                usermod -aG docker arm && echo -e "${GREEN}Adds user arm to docker user group${NC}"
```
mihawk90: Same as above.

scripts/installers/docker-setup.sh
```diff
@@ -26,51 +34,90 @@ done
...
+            dnf install -y curl lsscsi
+            ;;
+        rhel|centos|rocky|almalinux)
+            yum install -y curl lsscsi
+            ;;
+        alpine)
+            apk add --no-cache curl lsscsi
+            ;;
+        *)
+            echo "${RED}Unsupported Linux distro: $ID"
```
mihawk90: Missing color termination.

scripts/installers/docker-setup.sh
```diff
@@ -26,51 +34,90 @@ done
...
+        echo -e "${GREEN}arm group already exists, skipping...${NC}"
     fi
 
     # create arm user if it doesn't already exist
     if ! id arm >/dev/null 2>&1; then
         useradd -m arm -g arm
         passwd arm
     else
-        echo -e "${RED}arm user already exists, skipping...${NC}"
+        echo -e "${GREEN}arm user already exists, skipping...${NC}"
```
mihawk90, line 79: Should add a `usermod -aG arm arm` here. This makes sure that the user is always part of the group.

Above there is a check to create the `arm` *group* if it does not exist. And it creates the `arm` *user* if it does not exist.
However, if the user already existed it is never checked whether that user is also part of the group. `usermod` exits 0 if the group was already on the user, so this doesn't disrupt anything.

scripts/installers/docker-setup.sh
```diff
@@ -99,4 +146,5 @@ pull_image
 setup_mountpoints
 save_start_command
 
-echo -e "${RED}Installation complete. A template command to run the ARM container is located in: $(echo ~arm) ${NC}"
+ARM_HOME=$(eval echo ~arm)
```
mihawk90: Why the `eval` here? A subshell like this always returns the string:
```sh
❯ test_echo=$(echo ~tarulia)

❯ test_eval=$(eval echo ~tarulia)

❯ echo $test_echo
/home/tarulia

❯ echo $test_eval
/home/tarulia
```

mihawk90 (commented):
I've already raised my general concerns with installing tools (especially when users don't get to review before package installs due to `-y` flags), but regardless of that... some comments on the changes.

scripts/installers/docker-setup.sh
53845714nF: Good question—maybe shellceck got me to do it; I just made it ~arm now.

scripts/installers/docker-setup.sh
```diff
@@ -26,51 +37,87 @@ done
...
         groupadd arm
     else
-        echo -e "${RED}arm group already exists, skipping...${NC}"
+        echo -e "${GREEN}arm group already exists, skipping...${NC}"
     fi
 
     # create arm user if it doesn't already exist
     if ! id arm >/dev/null 2>&1; then
         useradd -m arm -g arm
+        usermod -aG arm arm
```
mihawk90: This seems to be in the wrong branch? This branch is executed when the user doesn't exist. The `useradd` already adds the group. The prior suggestion was to add the group to an *existing* user.

mihawk90 (commented):
.

Issue #1599, opened 2025-10-17
🐋 Cannot not get image to run "[ERROR]: ARM does not have permissions to /etc/arm/config using 1000:1000"


KayDox64:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Does this issue exist in the latest version?

- [x] I'm using the latest release

### Describe the bug?

Title, doesn't launch, here is the log.

### To Reproduce

pull docker image

run container

### Built from source files ?

- [ ] I Built the image from source

### What architecture are you seeing the problem on?

all

### Expected behavior?

for it to work?

### Relevant log output

```shell
*** Running /etc/my_init.d/10_syslog-ng.init...
Oct 17 02:37:22 d311a5db37ff syslog-ng[13]: syslog-ng starting up; version='3.35.1'
*** Running /etc/my_init.d/arm_user_files_setup.sh...
Updating arm group id 1000 to default (1000)...
usermod: no changes
Updating arm group id 1000 to default (1000)...
Adding arm user to 'render' group
Checking ownership of /home/arm
[OK]: ARM UID and GID set correctly, ARM has access to '/home/arm' using 1000:1000
Creating dir: /home/arm/media
Creating dir: /home/arm/media/completed
Creating dir: /home/arm/media/raw
Creating dir: /home/arm/media/movies
Creating dir: /home/arm/media/transcode
Creating dir: /home/arm/logs
Creating dir: /home/arm/logs/progress
Creating dir: /home/arm/db
Creating dir: /home/arm/music
Creating dir: /home/arm/.MakeMKV
Removing any link between music and Music
stat: cannot statx '/etc/arm/config': No such file or directory
Checking ownership of /etc/arm/config
---------------------------------------------
[ERROR]: ARM does not have permissions to /etc/arm/config using 1000:1000
 Check your user permissions and restart ARM. Folder permissions--> :
 ---------------------------------------------
 stat: cannot statx '/etc/arm/config': No such file or directory
 Oct 17 02:37:24 d311a5db37ff syslog-ng[13]: syslog-ng shutting down; version='3.35.1'
 *** /etc/my_init.d/arm_user_files_setup.sh failed with status 1
 *** Killing all processes...
Oct 17 02:40:25 d311a5db37ff syslog-ng[12]: syslog-ng shutting down; version='3.35.1'
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

1337-server:
[...]

Issue #1186, opened 2024-09-01
🐋 Starting fails trying to chown folders that are mounted by docker via NFS


btoconnor:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Does this issue exist in the latest version?

- [X] I'm using the latest release

### Describe the bug?

Starting the docker container fails with the error:
```
arm-rippers-1  | *** Running /etc/my_init.d/10_syslog-ng.init...
arm-rippers-1  | Sep  1 01:29:21 860e67934750 syslog-ng[12]: syslog-ng starting up; version='3.25.1'
arm-rippers-1  | *** Running /etc/my_init.d/arm_user_files_setup.sh...
arm-rippers-1  | Updating arm user id from 1000 to 1035...
arm-rippers-1  | Updating arm group id from 1000 to 100...
arm-rippers-1  | Adding arm user to 'render' group
arm-rippers-1  | Checking ownership of /home/arm
arm-rippers-1  | ---------------------------------------------
arm-rippers-1  | [ERROR]: ARM does not have permissions to /home/arm using 1035:100
arm-rippers-1  | Check your user permissions and restart ARM. Folder permissions--> 1035:1000
arm-rippers-1  | ---------------------------------------------
arm-rippers-1  | *** /etc/my_init.d/arm_user_files_setup.sh failed with status 1
arm-rippers-1  |
arm-rippers-1  | *** Killing all processes...
arm-rippers-1  | Sep  1 01:29:25 860e67934750 syslog-ng[12]: syslog-ng shutting down; version='3.25.1'
```

### To Reproduce

This is my configuration, with the synology ip redacted:

```yaml
version: "3.4"
services:
  arm-rippers:
    image: automaticrippingmachine/automatic-ripping-machine:latest
    restart: unless-stopped
    privileged: true
    volumes:
      - ./config:/etc/arm/config
      - armdata:/home/arm/media
      - arm-music:/home/arm/Music
      - ./logs:/home/arm/logs
    ports:
      - "8080:8080"
    environment:
      - ARM_UID=1035
      - ARM_GID=100

volumes:
  armdata:
    driver_opts:
      type: "nfs"
      o: "addr=synologyip,nolock,soft,rw"
      device: ":/volume1/arm-data/media"
  arm-music:
    driver_opts:
      type: "nfs"
      o: "addr=synologyip,nolock,soft,rw"
      device: ":/volume1/arm-data/Music"
```

[...]

Issue #822, opened 2023-04-10
🐋 setup scripts fail trying to chown folders that are mounted by docker 


gordallott:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Does this issue exist in the latest version?

- [X] I'm using the latest release

### Describe the bug?

the [arm_user_files_setup.sh](https://github.com/automatic-ripping-machine/automatic-ripping-machine/blob/main/scripts/docker/runit/arm_user_files_setup.sh#L46) script unnecessarily tries to chown directories, which will do nothing useful on most installs, and cause issues with anyone who has setup their nas to not allow random services to chown directories in the nas shares

if you have NFS setup to force specific groups so that every service you have is happy and does not encounter errors as random docker containers all have their own users/groups, these chown calls will fail causing the entire script to fail, causing the docker image to never enter a started state 

### To Reproduce

1. setup NFS with a mapall option to map all users to a specific user 
2. try and launch the ARM docker container
3. observe failure 

### Built from source files ?

- [ ] I Built the image from source

### What architecture are you seeing the problem on?

arm64

### Expected behavior?

the script should not attempt to chown folders that are expected to be mounted, it's always going to end in failure as different systems have different UAC requirements, it should just live within the defined permissions of the folders that are mounted. 

### Relevant log output

```shell
*** Running /etc/my_init.d/arm_user_files_setup.sh...
Updating arm group id 1000 to default (1000)...
usermod: no changes
Updating arm group id from 1000 to 1003...
Adding arm user to 'render' group
chown: changing ownership of '/home/arm/media': Operation not permitted
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

shitwolfymakes:
Thanks for pointing this out, I can see why it would be an issue. We do this to ensure that the files written by the container are not owned by root, otherwise you'd have to use sudo for every operation.

ARM specifically needs to be added to the render group, but that is something you'll have to do on the hosts end.

TheRouce:
Is there any update on this? 

The arm user on my host machine has all the privileges it needs to read and write to the nfs share but not to chown.

koeberlue:
I had the same problem when I was trying to use an SFTP mount. Here is what I did to make it run:

1. Inside my SFTP mount create a new folder, named `mymedia`
  1. So I had a folder `/path/to/mount/mymedia`
1.  `ls -n /path/to/mount/`
[...]

PR #1660, opened 2026-01-02
[BUGFIX] Fix file ownership of config files created in arm_user_files_setup.sh

### LINKED ISSUE #1639: In ARM UI, Ripper Settings changes are not saving
### Is there an existing issue for this?

- [x] I have searched the existing issues

### I have checked the wiki

- [x] I have searched the wiki for this problem

### Does this issue exist in the latest version?

### Describe the bug?

While in the ARM UI, navigating to Settings->Ripper Settings I was trying to update the HB_PRESET_DVD field from "HQ 720p30 Surround" to "Super HQ 720p30 Surround" and the HB_PRESET_BD field from "HQ 1080p30 Surround" to "Super HQ 1080p30 Surround".  At the bottom of the page I click Submit, but nothing appears to happen. 

I can navigate to another sub-tab (e.g. UI Settings) and back to the Ripper Settings tab and the fields appear to have updated with the changed values.  However, when I navigate to a parent tab (e.g. Home), then when I return to the Ripper Settings sub-tab, the fields have returned to the original default values.  

I've been able to repeat the issue using different browsers.  

### To Reproduce

1. In the ARM UI, click Settings
2. Click Ripper Settings
3. Click in the HB_PRESET_DVD: field and change the value
4. Click Submit at the bottom of the page
5. Click on the Home parent tab
6. Click on the Settings tab
7. Click on the Ripper Settings sub-tab
8. Observe the value in the HB_PRESET_DVD: field

### What OS are you seeing the problem on?

Other (please state in Anything else? section)

### ARM Release Version or Git hash (You can find this in the arm settings page)

Python version: 3.10.12
A.R.M version: 2.21.0
Current git version: 5132031

### Expected behavior?

- Upon clicking submit on the Ripper Settings sub-tab, receiving some kind of confirmation that the changes were committed
- Clicking on another parent tab and then back into the Settings->Ripper Settings sub-tab and seeing the fields with the updated values

### Relevant log output

```shell

```

### Anything else?

Environment: 
Base system=Ubuntu Desktop 24.04 LTS
ARM installation=Docker container

### Code of Conduct

dkreth:
# Description
When config files are copied from `/opt/arm/setup/` to `/etc/arm/config/`, they retain ownership (see [man page for cp](https://man7.org/linux/man-pages/man1/cp.1.html)). If the startup script runs as root, then the config files are created in `/etc/arm/config` and owned by root. This prevents the `arm` user from modifying the config files via the web UI, causing a 500 error when attempting to save settings in the ARM UI.

To fix this, we add a `chown` command after cp to ensure config files are owned by the `arm` user, matching the ownership requirements of the `/etc/arm/config` directory.

Fixes #1639 #1599 

## Type of change

# How Has This Been Tested?
- [x] Docker

# Checklist:

- [x] I have tested that my fix is effective

# Changelog:

- Add a `chown` command after cp to ensure config files are owned by the `arm` user


microtechno9000 (approved):
agree with change to remove depreciated cp command

dkreth:
I reverted the change to remove the deprecated cp command. The container uses coreutils 8.32, which doesn't support `--update=none` (introduced in coreutils 9.1).

dkreth:
Just tested this.

Script Output:
```
  Config not found! Creating config file: /etc/arm/config/arm.yaml
  Config not found! Creating config file: /etc/arm/config/apprise.yaml
```

File Ownership Result:
```
  -rw-r--r-- 1 arm  arm   4845 Jan 28 09:40 apprise.yaml
  -rwxr-xr-x 1 arm  arm  12649 Jan 28 09:40 arm.yaml
```
So both `arm.yaml` and `apprise.yaml` are owned by `arm:arm`.

dkreth:
## Problem
I just found [an issue](https://github.com/automatic-ripping-machine/automatic-ripping-machine/issues/822) that was fixed by [this](https://github.com/dkreth/automatic-ripping-machine/commit/a8a8b8f5bcf3bba7b025a6e1d6e01a0f1dd5267c) commit but may still be present according to [this issue](https://github.com/automatic-ripping-machine/automatic-ripping-machine/issues/1186) from August 2024.

So... I'm concerned that my fix that's being introducing in this PR (adding the chown command) will cause the GitHub issue above to regress/reoccur.

@microtechno9000 I had a couple ideas, but curious on your thoughts here.

## Option 1
I looked for a flag we could pass to the cp command, but didn't see one that fits our use case. The `--no-preserve=ownership` flag won't help here. The issue isn't copying ownership from the source, it's that cp creates the new file owned by whoever runs the command (root).

## Option 2
Another option would be to create the file as the `arm` user instead of `root`
```
su arm -c "cp --no-clobber '/opt/arm/setup/${conf}' '${thisConf}'"
```
I believe this would work because the script already runs as `root`, so `su arm` works directly, but @microtechno9000 can you verify my assumption here?

If so, I think this would be the cleanest solution:
  - No `chown` call = no NFS/SFTP permission failures
  - File created with correct ownership from the start
  - Works as long as `arm` user has write access to the directory (which I believe is already verified by `check_folder_ownership`)

### Option 3
Alternatively, the simplest option would be to just attempt the `cp` command and bailout if it fails. There's no chance at a reversion of issue #822, but it also will fail silently.
```
cp --no-clobber "/opt/arm/setup/${conf}" "${thisConf}"
chown arm:arm "${thisConf}" 2>/dev/null || true
```

microtechno9000:
> ## Problem
> 
> I just found [an issue](https://github.com/automatic-ripping-machine/automatic-ripping-machine/issues/822) that was fixed by [this](https://github.com/dkreth/automatic-ripping-machine/commit/a8a8b8f5bcf3bba7b025a6e1d6e01a0f1dd5267c) commit but may still be present according to [this issue](https://github.com/automatic-ripping-machine/automatic-ripping-machine/issues/1186) from August 2024.
> 
> So... I'm concerned that my fix that's being introducing in this PR (adding the chown command) will cause the GitHub issue above to regress/reoccur.
> 
> @microtechno9000 I had a couple ideas, but curious on your thoughts here.
> ## Option 1
> 
> I looked for a flag we could pass to the cp command, but didn't see one that fits our use case. The `--no-preserve=ownership` flag won't help here. The issue isn't copying ownership from the source, it's that cp creates the new file owned by whoever runs the command (root).
> ## Option 2
> 
> Another option would be to create the file as the `arm` user instead of `root`
> 
> ```
> su arm -c "cp --no-clobber '/opt/arm/setup/${conf}' '${thisConf}'"
> ```
> 
> I believe this would work because the script already runs as `root`, so `su arm` works directly, but @microtechno9000 can you verify my assumption here?
> 
> If so, I think this would be the cleanest solution:
> 
>     * No `chown` call = no NFS/SFTP permission failures
> 
>     * File created with correct ownership from the start
> 
>     * Works as long as `arm` user has write access to the directory (which I believe is already verified by `check_folder_ownership`)
> 
> 
> ### Option 3
> 
> Alternatively, the simplest option would be to just attempt the `cp` command and bailout if it fails. There's no chance at a reversion of issue #822, but it also will fail silently.
> 
> ```
> cp --no-clobber "/opt/arm/setup/${conf}" "${thisConf}"
> chown arm:arm "${thisConf}" 2>/dev/null || true
> ```

This has indeed been an issue that keeps coming up, in particular because ARM uses a privileged container due to needing access to CD/DVD/Bluray drives and GPU compute. There are ways around this, but not easily.

I think Option 2 is the best, it will avoid NFS permissions issues (NFS can be fun at the best of times) and avoid ARM changing permissions outside a container.

If you update the code to Option 2 I can merge in

microtechno9000 (changes requested):
pending change, update to option 2 and remove 'chown'

mihawk90:
Option 4 is just using the `install` command that was made for this purpose:

```sh
❯ man install

NAME
       install - copy files and set attributes

SYNOPSIS
       install [OPTION]... [-T] SOURCE DEST
       install [OPTION]... SOURCE... DIRECTORY
       install [OPTION]... -t DIRECTORY SOURCE...
       install [OPTION]... -d DIRECTORY...

…

       -g, --group=GROUP
              set group ownership, instead of process' current group

       -m, --mode=MODE
              set permission mode (as in chmod), instead of rwxr-xr-x

       -o, --owner=OWNER
              set ownership (super-user only)

…
```

On a side note, the `--no-clobber` feels rather redundant when the copy is only executed if the file isn't present to begin with, i.e. the script is already checking that the file is *not* present before copying, and then uses an option to not overwrite existing files?

microtechno9000 (changes requested):
A docker container shouldn't chown files that are created and exist outside the container. I understand the issue and prior versions of ARM changed the ownership of these files but ran in to a host of issues with users systems, due to docker trying to change ownership outside the container.

Suggest an alternative solution be investigated, i.e. `su -u arm`?

scripts/docker/runit/arm_user_files_setup.sh
```diff
@@ -87,7 +87,9 @@ for conf in $CONFS; do
   if [[ ! -f "${thisConf}" ]] ; then
     echo "Config not found! Creating config file: ${thisConf}"
     # Don't overwrite with defaults during reinstall
+    # Note: --no-clobber is deprecated in coreutils 9.1+, but container uses 8.32
     cp --no-clobber "/opt/arm/setup/${conf}" "${thisConf}"
+    chown arm:arm "${thisConf}"
```
microtechno9000, line 92: We cannot have arm change the ownership of files and folders from within a container, that is not good security practice and will lead to potential issues/breakages within the host OS

microtechno9000 (changes requested):

PR #1672, opened 2026-01-15
Rip only main feature (longest track) when MAXLENGTH unlimited

### LINKED ISSUE #1633: 🔧 Main Feature

Main Feature not found cancels rip

Ability to set Main Feature: true but in case of failure to find main feature causing rip to cancel, it would continue as if Main Feature was set to false and rip all tracks

The Current Options You Have
Since a fallback isn't possible, you have to choose one behavior over the other:
Option 1: Live with the "Rip All" Default (Recommended compromise)
The most reliable option is to keep MAINFEATURE: 0 (false). This ensures the machine never stops automatically.

    Pros: Ripping never cancels due to this error.
    Cons: You get every single track ripped (including FBI warnings, menus, etc.). You must manually go in later and delete the files you don't need, or configure your media server to ignore them.

Option 2: Use a Manual Workflow
If you want the machine to auto-identify correctly most of the time, keep MAINFEATURE: 1 (true).

    Pros: Most rips will be clean and automatic.
    Cons: When you encounter a problematic disc with copy protection, ARM will fail and eject the disc. You will then have to manually:
        Reconfigure arm.yaml to MAINFEATURE: 0.
        Restart the Docker container (docker restart arm-rippers-v3).
        Re-insert the disc.
        Change the setting back to 1 for the next disc.

Summary
There is no simple configuration change to automatically switch behavior upon failure. You must choose between stopping on failure (MAINFEATURE: 1) or ripping everything (MAINFEATURE: 0).

### Anything else?

_No response_

### Code of Conduct

paul-tastic:
## Summary

When `MAXLENGTH` is set to unlimited (>99998), this change makes ARM rip only the longest track (main feature) instead of all tracks meeting the `MINLENGTH` requirement.

This addresses a common user request to avoid ripping bonus features, trailers, and extras when using the MKV rip method.

## Changes

- Modified `makemkv_mkv()` in `arm/ripper/makemkv.py`
- When MAXLENGTH > 99998, queries for the longest track and rips only that
- Falls back to ripping all tracks if no track info is found

## Related Issues

Closes #1633 - This implements the fallback behavior requested, but goes further by making main-feature-only the default when MAXLENGTH is unlimited.

## Test Plan

- [x] Tested with DVD rip
- [x] Verified only longest track is ripped
- [x] Confirmed fallback works when track info unavailable

### DIFF
--- a/contrib/smart-identify/move-completed.sh
+++ b/contrib/smart-identify/move-completed.sh
@@ -1,0 +1,220 @@
+#!/bin/bash
+# ARM Post-Processing Script v3
+# Uses smart identification before moving to Radarr/Sonarr
+
+set -euo pipefail
+
+RAW="/mnt/media/rips/raw"
+COMPLETED="/mnt/media/rips/completed"
+MOVIES_DIR="/mnt/media/movies"
+TV_DIR="/mnt/media/tv"
+MUSIC_DIR="/mnt/media/music"
+LOG="/mnt/media/docker/arm/move-completed.log"
+SMART_ID="/mnt/media/docker/arm/smart-identify.sh"
+
+# API Keys
+RADARR_API="YOUR_RADARR_API_KEY"
+SONARR_API="YOUR_SONARR_API_KEY"
+RADARR_URL="http://localhost:7878"
+SONARR_URL="http://localhost:8989"
+
+# Email settings
+EMAIL_TO="your@email.com"
+
+log() {
+    echo "$(date '+%Y-%m-%d %H:%M:%S') - $1" >> "$LOG"
+}
+
+send_alert() {
+    local subject="$1"
+    local body="$2"
+    echo -e "$body" | mail -s "$subject" "$EMAIL_TO"
+    log "Email alert sent: $subject"
+}
+
+# Check if folder has finished ripping (no active makemkv process for this path)
+is_rip_complete() {
+    local folder="$1"
+    local folder_name
+    folder_name=$(basename "$folder")
+
+    # Check if MakeMKV is still writing to this folder
+    if docker exec arm pgrep -f "makemkv.*$folder_name" > /dev/null 2>&1; then
+        return 1
+    fi
+
+    # Check for video files
+    if ! find "$folder" -maxdepth 2 -type f \( -name "*.mkv" -o -name "*.mp4" \) 2>/dev/null | grep -q .; then
+        return 1
+    fi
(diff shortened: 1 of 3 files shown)


arm/ripper/makemkv.py
```diff
@@ -690,21 +690,27 @@ def makemkv_mkv(job, rawpath):
 
             # Setting rawpath to None to set the job as failed when returning to arm_ripper
             rawpath = None
-    # if no maximum length, process the whole disc in one command
+    # if no maximum length, rip only the main feature (longest track)
```
microtechno9000, line 693: Update the comments.json https://github.com/automatic-ripping-machine/automatic-ripping-machine/blob/5132031886f6c7d55343d600b56faa385f6265b8/arm/ui/comments.json#L24 to reflect this change

contrib/smart-identify/move-completed.sh
microtechno9000, line 1: interesting solution, ARM isn't to great at finding disc/movie names. If only there was a standard naming convention all studios used.
Some questions
1. how come this isn't in the main ARM ripper identification code?
2. how come there are specific movie look ups, seems that this will cause issues in the future as more titles need to be added?

microtechno9000:
> > This behaviour would break ripping TV series
> 
> Second this, and I am also somewhat confused why this would be shoehorned into the `MAXLENGTH` setting to begin with.
> There is already a `MAINFEATURE` setting for this very purpose:
> 
> https://github.com/automatic-ripping-machine/automatic-ripping-machine/blob/109db59bc8532cb9e5ad8dd26f5921cb0388f3ed/setup/arm.yaml#L250-L256
> 
> If this isn't working properly, then that should be the one to be fixed up, no?
> 
> ---
> 
> Also the `smart-identify` scripts seem to be unrelated to the purpose of this PR? They should IMO be PR'd separately, though I have concerns about their maintainability and feature scope to begin with, but that's a discussion for another day.

Agree, there are a few questions I have, hard coding values for discs is not a good idea.

microtechno9000:
Thank you for supporting the ARM project, but I am closing this PR

The proposed change will on review break some of how users are already using ARM.

The identification script provides additional ID support, but in a way that is limited and would require constant updates to support new discs, changes or major titles. If this is something you would still like to see, please look at making it a generic solution that everyone could use and raise a new PR with just this code, incorporated into the ARM ripper detection and identification code.

PR #1682, opened 2026-01-31
Add macOS UTM Docker installation guide

RayHollister:
This PR adds a new wiki page documenting how to install ARM using Docker inside a UTM virtual machine on macOS.

## Why this approach?

macOS cannot natively pass USB devices to Docker containers. This guide provides a workaround using UTM's QEMU backend to run Ubuntu Server in a VM, which then runs ARM in Docker with proper USB passthrough to optical drives.

## Key points covered:

- UTM installation and QEMU backend requirement (Apple Virtualization doesn't support USB passthrough)
- Ubuntu Server VM setup for both Apple Silicon and Intel Macs
- USB DVD drive passthrough configuration
- Docker and ARM container setup
- Optional VirtFS shared media library for direct output to macOS folders
- Troubleshooting common issues

## Testing

This guide has been tested and verified working on a Mac Mini M4 with an external USB DVD drive.


arm_wiki/Alternate-Install-macOS-UTM.md
```diff
@@ -0,0 +1,281 @@
+# macOS with UTM (Docker)
+
```
microtechno9000, line 2: Similar to other installs, add in the caution note
https://github.com/automatic-ripping-machine/automatic-ripping-machine/blob/109db59bc8532cb9e5ad8dd26f5921cb0388f3ed/arm_wiki/Alternate-Install-Debian.md?plain=1#L5

RayHollister: Thanks for the review! I've made both changes:

1. Added the `> [!CAUTION]` block after the title, matching the format from the Debian install page. I adjusted the wording slightly since this guide *does* use Docker (just inside a VM): "This installation method was developed for macOS users who cannot run Docker natively with USB passthrough."

2. Added `[macOS with UTM (Docker)](Alternate-Install-macOS-UTM)` to `_Sidebar.md` under the Alternate Installations section, after TrueNAS.

microtechno9000 (approved):
Thanks for adding a guide for mac :)

PR #1739, opened 2026-03-25
feat: always send per-track naming manifest to transcoder

uprightbass360:
## Summary
- Remove `multi_title` guard on track metadata in webhook payload
- ARM always provides rendered filenames for every track — transcoder uses them directly
- Single-title jobs now include the same track manifest as multi-title
- ARM is the single source of truth for all output naming

## Test plan
- [ ] Verify single-title movie rip sends track manifest in webhook payload
- [ ] Verify multi-title series rip still sends track manifest with per-track names
- [ ] Verify transcoder uses ARM-provided filenames for output
- [ ] `pytest test/test_coverage_gaps_2.py::TestBuildWebhookPayload -v` — all 7 tests pass


uprightbass360:
Opened against wrong repo — closing

PR #1741, opened 2026-03-30
fix: async blocking audit — prevent NFS stalls in API

uprightbass360:
## Summary

- **Convert async handlers to sync** in UI backend to prevent event loop blocking (file browser + other handlers)
- **Cached disk usage via subprocess probe** (`disk_usage_cache.py`) — background refresh every 30s with 5s timeout, so `statvfs()` on stalled NFS never blocks API threads
- **Protect `/system/paths` endpoint** — `os.path.exists()` and `os.access()` now go through the same cached subprocess probe instead of blocking directly
- **Rsync copy** for NFS-safe file transfer during rip completion
- **CodeQL fixes** — replace polynomial regex patterns in `arm_matcher.py`

## Details

NFS mounts can enter kernel D-state during large file operations, causing `statvfs()`, `os.path.exists()`, and `os.access()` to block indefinitely. Threads stuck in D-state cannot be interrupted, eventually exhausting the entire threadpool and making the API unresponsive.

The fix uses subprocess-based probes with timeouts. If a subprocess stalls on NFS, it can be abandoned without consuming a threadpool slot. All 6 configured paths (RAW_PATH, TRANSCODE_PATH, COMPLETED_PATH, LOGPATH, DBFILE, INSTALLPATH) are pre-registered at startup for background refresh.

## Test plan

- [x] All 1686 tests pass
- [ ] Deploy to hifi-server and verify `/system/paths` returns cached results
- [ ] Verify dashboard disk usage stats still update correctly
- [ ] Simulate NFS stall and confirm API remains responsive

### DIFF
--- a/arm/api/v1/system.py
+++ b/arm/api/v1/system.py
@@ -172,7 +172,14 @@ def get_version():
 
 @router.get('/system/paths')
 def get_paths():
-    """Check existence and writability of configured ARM paths."""
+    """Check existence and writability of configured ARM paths.
+
+    Reads from the disk-usage cache to avoid blocking on stale NFS mounts.
+    Falls back to direct checks only for paths not yet in the cache (e.g.
+    DBFILE, LOGPATH which are local and won't stall).
+    """
+    from arm.services.disk_usage_cache import get_path_status, register_paths as _reg
+
     path_keys = [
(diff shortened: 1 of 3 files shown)


uprightbass360:
Opened against wrong repo by mistake, closing.

PR #1791, opened 2026-07-11
[BUGFIX] Active Rips card renders blank while a rip is in progress

Sawtaytoes:
## Problem

In my testing, the home page **Active Rips** card renders blank while a rip is in progress. The cause is that `GET /json?mode=joblist` 500s with:

```
'NoneType' object has no attribute 'group'
```

## Root cause

In `arm/ui/json_api.py`, `process_makemkv_logfile()` computes an ETA inside `if job_progress_status is not None:` and unconditionally calls:

```python
job_start_time = int(job_batch_info.group(1))
```

`job_batch_info` is parsed from the `<job_id>.log.batchinfo` file — but that file is only written **later in the same function**, inside the `if job_stage_index is not None:` block. On the first poll for a job the file doesn't exist yet, so `job_batch_info` is `None` and `.group(1)` raises, which 500s the whole endpoint and blanks the card.

Because the crash happens **before** the code that would create the batch-info file, it never bootstraps — so for affected rips the card stays blank for the entire job. (A `makemkv` progress log with `PRGV` lines but no seeded `BINF` reliably reproduces it.)

## Fix

Guard the ETA calculation on `job_batch_info` being present (and on progress having advanced past 0, which also avoids a `ZeroDivisionError` in `(elapsed_time * 100) / float(job.progress)`), reporting an `Unknown` ETA until batch info is available. The progress **percentage** itself comes from `PRGV` and is unaffected — so the % bar now renders immediately instead of the endpoint 500ing.

Minimal and behaviour-preserving: once the batch-info file exists, the ETA path runs exactly as before.

## Testing

- `python -m py_compile arm/ui/json_api.py` ✓
- Reproduced against a live rip (makemkv writing `progress/<id>.log` with `PRGV` lines, no `.batchinfo` yet): before, `/json?mode=joblist` returned the `NoneType` error and the card was blank; after, it returns the job list with a live percentage and `eta: Unknown` until batch info appears.


Sawtaytoes:
Looks like the Greeting one failed on some other PRs too.

microtechno9000:
> Looks like the Greeting one failed on some other PRs too.
> 

That's OK, the greeting check keeps failing and needs fixing

microtechno9000 (approved):

PR #1799, opened 2026-07-27
[BUGFIX] Don't search the metadata provider for the "not identified" placeholder

### LINKED ISSUE #1798: Unlabelled DVDs are confidently misidentified as "Prefiro não ser identificada" (2018)
# Unlabelled DVDs are confidently misidentified as "Prefiro não ser identificada" (2018)

When a DVD's volume label cannot be read, ARM substitutes the placeholder string
`"not identified"` and then **searches the metadata provider for that string**. TMDB
returns a real 2018 Brazilian film, *Prefiro não ser identificada* ("I'd rather not be
identified"), and `update_job()` writes it into the job as a successful match.

The result is not a failed identification — it is a **confident wrong one**. The job is
flagged `hasnicetitle: True`, and the rip is filed into the media library under the
wrong title and year.

## Environment

| | |
|---|---|
| ARM version | 2.24.2 |
| Commit | `4528fd522857dc43758cc0cfb351dd29be64858f` |
| Image | `automaticrippingmachine/automatic-ripping-machine:latest` |
| Image digest | `sha256:565c70e9cf0fdf3be60dd5984abca239ebae12b5ca108f3d9e6b87a772fedfa0` |
| Python | 3.10.12 |
| Platform | x86_64, Docker, `privileged: true` |
| `METADATA_PROVIDER` | `tmdb` |
| `RIPMETHOD` / `RIPMETHOD_DVD` | `mkv` / `mkv` |

Reproduced on two independent jobs (job 3 and job 6) with two different discs.

**Per CONTRIBUTING:** MakeMKV and HandBrake are not implicated. This occurs during disc
identification, before either program is invoked — the log below shows the wrong title
assigned at 11:49:08, while the MakeMKV rip does not begin until 11:50:16. That rip then
completed successfully (4.18 GB, no read errors, no corruption reported), so the disc and
the ripping path are both healthy. Only the title is wrong.

### Branches affected

The unguarded `metadata_selector()` is present on every active branch:

| branch | path | placeholder assigned | dead guard |
|---|---|---|---|
| `main` (2.24.2) | `arm/ripper/identify.py` | line 157 | line 243 |
| `v2_devel` | `arm/ripper/identify.py` | line 109 | line 173 |
| `3.0_devel` | `arm/ripper/main/identify.py` | line 145 | line 211 |

On `3.0_devel` the function has been moved but not changed — `search_results = None`
still follows the docstring directly, with no validation of `title`.

## Impact

Severity comes from the failure being **silent and confident** rather than loud:

* `hasnicetitle: True` is set, so the job presents as correctly identified in the UI
  and in `arm.db`.
* The output is filed into the configured `COMPLETED_PATH` under the wrong title,
  e.g. `<COMPLETED_PATH>/movies/Prefiro-não-ser-identificada (2018)/`.
* A media server scanning that directory will attach the wrong poster, synopsis and
  metadata to the film.
* Every unlabelled disc collides into the *same* wrong folder, so successive rips
  overwrite or pile up under one incorrect title.

An honest "unidentified" outcome would be strictly better, and the code already
intends that — see *Existing guard* below.

## Steps to reproduce

1. Set `METADATA_PROVIDER: "tmdb"` with a valid `TMDB_API_KEY`.
2. Insert a DVD whose volume label ARM does not read. This happens with genuinely
   unlabelled discs, and also — as in our case — when ARM wins a race against udev and
   reads the device before the filesystem probe populates `ID_FS_LABEL`. Symptom in the
   container log: `Starting ARM for unknown disc type on srN`.
3. Let identification run.

**Expected:** the disc is treated as unidentified and filed under the unidentified path
for manual naming.

**Actual:** the job is titled *Prefiro-não-ser-identificada*, year 2018,
`hasnicetitle: True`.

## Root cause

All line numbers are from `arm/ripper/identify.py` at the commit above.

```python
150  def identify_dvd(job):
155      # Some older DVDs aren't actually labelled
156      if not job.label or job.label == "":
157          job.label = "not identified"          # placeholder assigned
158      try:
159          crc64 = pydvdid.compute(str(job.mountpoint))
160          dvd_title = f"{job.label}_{crc64}"    # CRC-suffixed variant
...
188          dvd_title = str(job.label)            # (pydvdid failure branch)
...
195      dvd_title = job.label.replace("16x9", "") # <-- unconditionally discards
                                                  #     both of the above
...
203          dvd_info_xml = metadata_selector(job, dvd_title, year)   # searches
                                                  #     "not identified"
205          identify_loop(job, dvd_info_xml, dvd_title, year)
```

Line 195 resets `dvd_title` to the bare label regardless of which branch above ran, so
the placeholder reaches `metadata_selector()` at line 203 on every path.
`metadata_selector()` (line 292) applies no validation before dispatching:

```python
303      search_results = None
304      if cfg.arm_config['METADATA_PROVIDER'].lower() == "tmdb":
306          search_results = ui_utils.tmdb_search(title, year)
307          if search_results is not None:
308              update_job(job, search_results)   # writes the bogus match into the job
```

### Existing guard, and why it does not fire

`get_video_details()` already checks for exactly this placeholder:

```python
243      if title == "not identified" or title is None or title == "":
244          logging.info("Disc couldn't be identified")
245          return
```

The intent is right, but the check is downstream of the damage. It reads `job.title`,
and by the time it runs, `update_job()` at line 308 has already replaced `job.title`
with `"Prefiro-não-ser-identificada"`. The guard compares the *bogus match* against the
placeholder, does not match, and allows processing to continue.

## Evidence

Container log (job 6):

```
11:48:43  [ARM] Starting ARM for unknown disc type on sr1
11:48:54  INFO: Starting Disc identification
11:48:58  INFO: Disc identified as video
11:49:08  INFO: Disc title Post ident -  title:Prefiro-não-ser-identificada year:2018 \
                video_type:movie disctype: dvd
11:50:09  INFO: title: Prefiro-não-ser-identificada
11:50:09  INFO: hasnicetitle: True
11:50:09  INFO: label: not identified
```

`arm.db`, two independent jobs converging on the same wrong title:

```
job 3: label='not identified' title='Prefiro-não-ser-identificada' year=2018 imdb=None
       path=<COMPLETED_PATH>/movies/Prefiro-não-ser-identificada (2018)
job 6: label='not identified' title='Prefiro-não-ser-identificada' year=2018 imdb=None
       path=<COMPLETED_PATH>/movies/Prefiro-não-ser-identificada (2018)_<JOBID>
```

Both jobs were the same disc — a commercial DVD carrying a short studio abbreviation as
its volume label, which udev did eventually report. ARM simply read it too early.

### The label was readable — ARM read it too early

The same disc, in the same session, produced both of these log files:

```
11:48:43  [ARM] Starting ARM for unknown disc type on sr1   -> job 6, label "not identified"
11:51:46  <LABEL>.log: "Device /dev/sr1: Job (6) status 'ripping'"
```

The second log is *named after the disc's real label* because by 11:51:46 — three minutes
after the job started — a later udev event carried that label without difficulty. The
disc was labelled the whole time; ARM simply queried before udev's filesystem probe had
populated `ID_FS_LABEL`. On this hardware the probe took roughly four minutes to complete
on a cold disc.

This is worth separating from the misidentification itself. The race is *how users reach*
the bug; the confident wrong answer is the bug. Fixing the race would reduce how often
this is hit but would not stop an genuinely unlabelled disc from being filed as a
Brazilian drama.

## Proposed fix

Refuse the search at the single chokepoint every caller passes through. All seven
`metadata_selector()` call sites within `identify.py` (lines 203, 341, 348, 353, 366,
384, 391) are covered by this one guard.

Patch file: `no-placeholder-search.patch` (applies with `patch --forward --strip=0`).

```diff
@@ -300,6 +300,16 @@
 
     :return: json/dict object or None
     """
+    # A disc whose label could not be read is given the placeholder
+    # "not identified" in identify_dvd(). Searching for that string finds real
+    # films with unhelpful names -- TMDB matches a 2018 Brazilian film titled
+    # "Prefiro nao ser identificada" -- and update_job() then overwrites
+    # job.title with it. The guard in get_video_details() cannot catch this,
+    # because by the time it runs job.title is the bogus match rather than the
+    # placeholder. Refuse the search here instead, where every caller passes.
+    if title is None or str(title).strip().lower() in ("", "not identified"):
+        logging.info(f"Not searching for placeholder title: {title!r}")
+        return None
     search_results = None
     if cfg.arm_config['METADATA_PROVIDER'].lower() == "tmdb":
         logging.debug("provider tmdb")
```

With this applied, `metadata_selector()` returns `None`, `identify_loop()` receives no
response, and the failsafe at lines 210–212 leaves `job.title` as the label — the
unidentified outcome the existing guard was written to produce.

## Notes and alternatives

* **A named constant would be preferable.** The string `"not identified"` is currently
  written at line 157 and compared at line 243; the patch adds a third site. A module
  constant (e.g. `UNIDENTIFIED_LABEL`) referenced in all three would be more robust. We
  kept the patch minimal to keep the diff reviewable, but would happily submit that
  version instead if preferred.
* **The placeholder is arguably the deeper bug.** Assigning human-readable prose to
  `job.label` makes it indistinguishable from a real label downstream. `None` or a
  sentinel would prevent this class of problem outright, at the cost of a wider change.
* **`hasnicetitle` should probably not be set** on a match derived from a placeholder,
  independent of this fix.
* **A related race is worth noting.** ARM triggers on the udev change event but reads
  `ID_FS_LABEL` before the filesystem probe completes on slow-spinning discs, which is
  what produced an empty label here. The disc *was* labelled. That is a separate issue
  from the misidentification, but it is the most likely way users encounter this one.

## Submission

**Yes — I am willing and able to submit this as a pull request.** The fix is committed
locally as a single 10-line change with a full explanation in the commit message, ready
to push to a fork.

### A note on CONTRIBUTING.md

The contributing guide asks for bug fixes against `v2_fixes` and features against
`v2.x_dev`. Neither branch exists on the remote any more:

```
$ git ls-remote --heads origin v2_fixes v2.x_dev
(no output)
```

The live branches are `main` (default, where `2.24.2` is tagged), `master`, `v2_devel`,
`v2_master` and `3.0_devel`. I have therefore branched from `main` — **please tell me if
this should be retargeted**, and I will rebase.

The guide also asks contributors to confirm the Travis CI build passes. CI appears to be
GitHub Actions now (`.github/workflows/main.yml` — CodeQL plus UI tests on Python
3.9–3.12); there is no Travis configuration in the tree. This change is stdlib-only and
introduces no version-specific behaviour. `flake8` passes: the longest added line is 78
characters against the 160-character limit in `setup.cfg`.

Both points are offered as a courtesy — the guide may simply predate a reorganisation,
and updating it would spare the next contributor the same detour.

### Scope

Kept deliberately minimal, per the guidance about separate pull requests. The
`UNIDENTIFIED_LABEL` constant suggested above, and the udev label race, are each better
as their own change; I am happy to follow up with either.

One unrelated issue was noticed in the same log and is **not** included here — `lsdvd`
output failing to parse at `identify.py:220`, which is caught at line 226 and only
disables Track 99 detection. I will file that separately if it is not already known.

## Credit

Found while building a DVD archiving pipeline on top of ARM.

bquandt:
Fixes #1798.

When a DVD's volume label cannot be read, `identify_dvd()` substitutes the placeholder string `"not identified"` and then searches the metadata provider for it. TMDB returns a real 2018 Brazilian film, and `update_job()` writes it into the job — so the result is not a failed identification but a confident wrong one, flagged `hasnicetitle: True` and filed into the library under the wrong title.

The existing guard in `get_video_details()` cannot catch this: by the time it runs, `job.title` has already been replaced with the bogus match, so it no longer equals the placeholder.

This refuses the search in `metadata_selector()` instead, which all seven call sites in `identify.py` pass through. The function returns `None`, `identify_loop()` gets no response, and the failsafe leaves `job.title` as the label — the unidentified outcome the existing guard was written to produce.

Verified on ARM 2.24.2 (`4528fd5`): `py_compile` passes, and flake8 is clean (longest added line is 78 characters against the 160-character limit in `setup.cfg`).

The same unguarded function is present on `v2_devel` and on `3.0_devel` (`arm/ripper/main/identify.py`), if you would like the fix applied there too.

One note: CONTRIBUTING.md asks for bug fixes against `v2_fixes` and features against `v2.x_dev`, but neither branch exists on the remote any more, so I have targeted `main`. Happy to rebase if that is not the right target — and updating the guide would spare the next contributor the same detour.


microtechno9000 (approved):
No issues with the change. No need to update the v2_devel or other branches. they are now obsolete
the v3_dev branch is in work and will be overwritten with a refactor

